PR #4, opened 2015-01-07
Make client pip installable

cranberyxl:
**Note**: This is a backwards compatibility break for anyone that is using this library directly via git

This PR makes the library installable via pip, initially as noted in the changed `README.md`, using the github url

To make this live happily with the rest of the pip installed modules, the main code was moved into a `clarifai` package. All documentation and samples are updated.


adamberenzweig:
Andy, thanks for this!  I just submitted a similar change, with a small difference (it makes the package path clarifai.client, not simply clarifai).

PR #8, opened 2015-07-06
python3 compatible sample

robertwenquan:
Last night there was a bug filed with Exception generated while importing clarifai.client with Python3. Here is the preliminary fix for it. We should fix it for the Python3 users.


adamberenzweig:
LGTM!

PR #15, opened 2015-10-08
Don't pass 'default' model

michalwols:


clarifai/client/client.py
```diff
@@ -136,7 +136,7 @@ def __init__(self, app_id=None, app_secret=None, base_url='https://api.clarifai.
     self.api_info = None
 
   def set_model(self, model):
-    self._model = self._sanitize_param(model, 'default')
+    self._model = self._sanitize_param(model)
```
adamberenzweig: If I read the current code right, we're already not sending an explicit model param in the default case.  self._model is None, and _sanitize_param returns None when it gets None.  I think this change isn't necessary.

clarifai/client/client.py
```diff
@@ -592,7 +592,7 @@ def _setup_multi_data(self, ops, num_cases, model=None, local_ids=None, meta=Non
     """ Setup the data dict to POST to the server. """
     data =  {'op': ','.join(ops)}
     if model:  # use the variable passed into method
-      data['model'] = self._sanitize_param(model, 'default')
+      data['model'] = self._sanitize_param(model)
```
adamberenzweig: In case the param has bad unicode, the client should just raise an error, rather than silently switching to a different model.  And i'm not 100% convinced that the server will treat an empty string the same as not sending a model param.

clarifai/client/client.py
michalwols: Yeah you're right. Either way, we still shouldn't convert the error to 'default'.

adamberenzweig: I think just letting the unicode error get raised is appropriate.

PR #17, opened 2015-10-27
refer to the server for languages

zeiler:
Let's let the server decide on what's supported or not instead of baking in options in the client.


vinayan3:
LGTM

PR #21, opened 2015-12-28
Add 'six' to requirements.

solongordon:
A user reported issues importing the client because he didn't have the
'six' module installed. I added this to the install requirements in
setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ setup(
     url='https://github.com/clarifai/clarifai_py',
     author_email='support@clarifai.com',
     version='0.2',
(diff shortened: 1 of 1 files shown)


solongordon:
@jimreesman for review

jimreesman:
i didn't check if other requirements have crept in - would you pls double-check? tx. then it's gtg.

solongordon:
Good question, I'll check.

solongordon:
Nope, looks like it's just six. I don't have write access anymore so I'll leave the merge to you.

PR #26, opened 2016-07-04
Adding Color to list of endpoints.

hammygoonan:


clarifai/client/client.py
```diff
@@ -739,3 +740,82 @@ def _base64_encoded_data_op(self, data, op):
     headers = self._get_json_headers()
     response = self._get_json_response(url, data=data, headers=headers)
     return self._parse_response(response)
+
+  def tag_color(self, files, model=None, local_ids=None, meta=None,
```
zeiler: Maybe just color(). 

"model", "select_classes", and "language" are not supported so please remove from args and the docs below. You can juts pass in their defaults of None to the _multi_data_op call for now.

clarifai/client/client.py
```diff
@@ -739,3 +740,82 @@ def _base64_encoded_data_op(self, data, op):
...
+                select_classes=None, language=None):
+    """ Autotag a single data file from an open file object or multiples data
```
zeiler: Please update doc to mention color instead of tag.

clarifai/client/client.py
```diff
@@ -739,3 +740,82 @@ def _base64_encoded_data_op(self, data, op):
...
+      clarifai_api = ClarifaiApi()
+      clarifai_api.tag_color([open('/path/to/local/image.jpeg'),
+                              open('/path/to/local/image2.jpeg')])
+    """
+    return self._multi_data_op(files, ['color'], model=model,
+                               local_ids=local_ids, meta=meta,
+                               select_classes=select_classes,
+                               language=language)
+
+  def tag_color_urls(self, urls, model=None, local_ids=None, meta=None,
```
zeiler: Same fixes here as above.

zeiler:
Thanks for the PR, just a few simple clean-ups of args and docs then we can merge!

hammygoonan:
Hey @zeiler how's that now?

clarifai/client/client.py
```diff
@@ -739,3 +740,75 @@ def _base64_encoded_data_op(self, data, op):
...
+      meta: a string of any extra information to accompany the request. This
+            has to be a string, so if passing structured data, pass a
+            json.dumps(meta) string.
+
+    Returns:
+      results: an API reponse including the generated tags. See the docs at
+      https://developer.clarifai.com/docs/ for more detais.
+
+    ## Example:
+    ```
```
robertwenquan: This is a Markdown syntax, probably we don't want to keep it in the python code.

clarifai/client/client.py
```diff
@@ -739,3 +740,75 @@ def _base64_encoded_data_op(self, data, op):
...
+                    Ex: "dog,cat,tree,car,boat"
+    language: set the default language using it's two letter (with options -XX
+              variant) ISO 639-1 code to use for all requests.
+
+    Returns:
+      results: an API reponse including the generated tags. See the docs at
+      https://developer.clarifai.com/docs/ for more detais.
+
+    ## Example:
+    ```
```
robertwenquan: here as well

PR #27, opened 2016-07-07
bumping minor version number

hammygoonan:
Adding color is new functionality without breaking backwards compatibility.

I also amended the doc string to remove Markdown.


robertwenquan:
Thanks!

PR #38, opened 2016-11-07
Fix breaking print in the v2.0.11 release for python 3.X

Stegoo:
Thank you for providing a python API :). I wanted to play around with the custom training API using Python 3.5 but could not get the 2.0.11 to work because of this line (I guess a printing debug left over) : https://github.com/Clarifai/clarifai-python/blob/v2.0.11/clarifai/rest/client.py#L2064

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Installation
 ---------------------
 Unzip the package to somewhere then run setup.py to install:
 ```
-pip install clarifai==2.0.10
+pip install clarifai==2.0.11
 ```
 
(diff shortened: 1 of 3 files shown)


robertwenquan:
@Stegoo Sorry for the inconvenience and thanks for sending and fixing the broken issue. 

We will make sure the github is always synced up with the released version, and it's always working on Python3. Currently we developed and tested heavier on python2 and we will make sure we put more effort on python3 as well.

PR #49, opened 2017-03-07
Add most recent Python versions in Travis CI

SylvainDe:
Add more recent Python versions including development branches and nightly build.

The motivation came from reading @brettcannon's article : https://snarky.ca/how-to-use-your-project-travis-to-help-test-python-itself/ . Trying to activate the newest Python versions on CI jobs is in most case a win-win situation: if everything works fine, there is nothing to worry about. If an issue is spotted, it is good to know about it to fix it on your side or to open a bug on CPython ( https://bugs.python.org/ ).

Also, if a failures is spotted, you can use the allow_failures option in your matrix build (more information about this in the link above).

Also, nightly and 3.7-dev may be a bit too recent/unstable so I can remove then if need be.

More information about how this PR happened on https://github.com/SylvainDe/CIthon .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ language: python
 python:
   - "2.7"
   - "3.5"
+  - "3.6"
+  - "3.6-dev"
+  - "3.7-dev"
+  - "nightly"
 # command to install dependencies
 install: "pip install -r requirements.txt"
 # command to run tests


robertwenquan:
Hi @SylvainDe 

Thanks for making the Pull Request and adding more python versions to the Travis CI.
It's intentional to have only python 2.7 and python 3.5 as we only spent minimum effort to support the client with python3.x 

Getting more recent versions supported may need some time to plan. I have that in my mind and have then appended to my TODO list. But it's not gonna happen in the near few weeks.

robertwenquan:
Thanks. I've added all of them.

PR #56, opened 2017-06-01
fix the typo on the home page of the clarifai python docs

rigeljs:


robertwenquan:
@rigeljs  Thanks for correcting it. This file is actually a copy from https://github.com/Clarifai/private-clarifai-python/blob/master/docs/index.rst I've updated the source one.

PR #57, opened 2017-06-02
Fix broken links in README

soupault:


robertwenquan:
@soupault sorry for the broken link. Actually the link is not broken. It's the missing tag in github that leads to the missing doc with the correct version tag. It has been fixed.

PR #100, opened 2019-03-01
Hard code protobuf version to 2.6.1

reverson:
Protobuf 3.7.0 was released and changed the init signature for _Parser, which is subclassed for _CustomParser.

This is a work around until we can get a more finalized solution in place.


reverson:
@rok-povsic is going to fix this with a more permanent solution.

PR #106, opened 2019-12-13
[TT-390]: Removed feedback endpoints from client

zeiler:
Removing these deprecated endpoints.


zeiler:
Removed for private one instead.

PR #107, opened 2020-01-02
Update to latest internal version

rok-povsic:


sonar-project.properties
```diff
@@ -0,0 +1,14 @@
+# sonarqube server url
+sonar.host.url=https://sonarqube.cf.ai
```
zeiler: @robertwenquan please review to make sure there are no secrets in this PR please. This is merging all private repo into public for good.

scripts/sync_repo.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
```
zeiler: Do we need this file?

rok-povsic: No, good catch. Removed.

sonar-project.properties
robertatclarifai: there is no secrets in this file. But this file might not be good for public consumption. 
The `sonarqube.cf.ai` itself should be internal only

tests/requirements.txt
```diff
@@ -10,4 +10,4 @@ pytest-cov==2.6.0
 pytest-xdist==1.23.2
 responses>=0.5
 retrying==1.3.3
-yapf==0.22.0
+yapf==0.19.0
```
robertatclarifai: why this is downgraded?

zeiler: To match main repo

PR #109, opened 2020-02-06
[TT-544]: remove Face object and facedetect* model types

zeiler:
These will be deprecated in coming week. See https://docs.clarifai.com/product-updates/upcoming-api-changes for more details on the Face proto deprecation and the `facedetect*` model type cleanup.

### DIFF
--- a/tests/unit_tests/test_model.py
+++ b/tests/unit_tests/test_model.py
@@ -84,7 +84,7 @@ def test_create_model(mock_http_client):  # type: (mock.Mock) -> None
       "output_config": {}
     }
   }
-} 
+}
       """)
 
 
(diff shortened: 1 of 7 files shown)


tests/unit_tests/test_model.py
```diff
@@ -185,8 +185,8 @@ def test_get_all_models(mock_http_client):  # type: (mock.Mock) -> None
       "app_id": "main",
       "output_info": {
         "message": "Show output_info with: GET /models/{model_id}/output_info",
-        "type": "facedetect",
-        "type_ext": "facedetect"
+        "type": "detect",
+        "type_ext": "detect"
```
rok-povsic: Do a `detect` type & type_ext really exist? I only see `detection`.

In any case, it's a unit test so it's not very important what the exact value is.

zeiler: true, yeah they will :) going to clean that up too eventually in the API. Not sure why we return `detection` but then `detect-embed` and `detect-concept`.

PR #112, opened 2020-03-13
Deploy to PyPi and Github Releases via Travis

rok-povsic:
Previously these steps had to be done manually.

On next deployment make sure all works fine, otherwise adjust.


zeiler:
Nice, what release version will it get from this?

rok-povsic:
@zeiler It should be [version from setup.py](https://github.com/Clarifai/clarifai-python/blob/master/setup.py#L16).

PR #113, opened 2020-03-24
[TT-791]: remove public workflows endpoint from client

zeiler:
Remove public_workflows endpoint. Users should rely on docs to get any public workflows or see them in the list when creating applications in Portal. 

image


HarmitMinhas96:
Do we want to keep the filter though for user to list workflows similar to list models? Where we filter out the non public ones from the list call?

zeiler:
no because the ListWorkflows endpoint already only lists those in your app which is the correct future behavrior. ListModels and all the others will be changed to that in near future too.

PR #115, opened 2020-04-09
Use environment markers for conditional dependencies

joshua-s:
See: https://www.python.org/dev/peps/pep-0508/#environment-markers

This fixes package installation with package managers like Poetry.


joshua-s:
@zeiler @rok-povsic could you please review this

joshua-s:
Is this still maintained?

PR #135, opened 2023-07-26
[DEVX:109] Added Authentication setup

sainivedh:
## what
 - To support backward compatibility, authentication (`clarifai.client.auth`) has same files that as current python utils
 - Users are expected to set `PAT` as env variable


sanjaychelliah (commented):
LGTM

clarifai/client/auth/abc.py
```diff
@@ -0,0 +1,23 @@
+import abc
+import grpc
+
+from clarifai_grpc.grpc.api import service_pb2_grpc
+
+
+class V2Stub(abc.ABC):
+  '''Abstract base class of clarifai api rpc client stubs'''
```
steph-en-m, line 8: Try to make consistent use of quotation marks throughout the docstrings. We should probably stick to """triple double quotes""" around docstrings

clarifai/client/auth/helper.py
```diff
@@ -0,0 +1,334 @@
...
+    self._pat = pat
+    self._token = token
+
+    self._base = https_cache(base_https_cache, base)
+    self._ui = https_cache(ui_https_cache, ui)
+
+  @classmethod
+  def from_streamlit(cls, st: Any, fallback_to_envvars: bool = True) -> "ClarifaiAuthHelper":
+    """This is a convenient method to check the query params from streamlit for the required
+        parameters for authentication, and then optional fallback to checking environment variables as
```
steph-en-m, line 111: For docstring convention, should we follow this below for one liner docstrings and multi-line docstrings?

One-liner:
```python3
def my_func(*args):
    """Short one liner docstrings."""
    pass
```

Multi-line docstrings:
```python3
def my_func(*args):
    """Function summary line (should be a one liner).
    
    Further function description.
    Args:
      arg_a (type):
          arg_a description.
    Returns:
      None
    """
    pass
```
The function summary line followed by a blank line below may come handy for docs incase a docstring parsers are used.

clarifai/client/base_auth.py
```diff
@@ -0,0 +1,30 @@
...
+        userDataObject (UserAppIDSet): The protobuf object representing user and app IDs.
+        base (str): The base URL for the API endpoint.
+    """
+
+  def __init__(self, **kwargs):
+    self.auth_helper = ClarifaiAuthHelper(**kwargs)
+    self.STUB = create_stub(self.auth_helper)
+    self.metadata = self.auth_helper.metadata
+    self.userDataObject = self.auth_helper.get_user_app_id_proto()
+    self.base = self.auth_helper.base
```
steph-en-m, line 30: Is the `BaseAuth` class really needed? 
It looks like everything it provides is already given in `ClarifaiAuthHelper()`

clarifai/client/user_api.py
```diff
@@ -0,0 +1,48 @@
+from clarifai_grpc.grpc.api import resources_pb2, service_pb2  # noqa: F401
+
+from clarifai.client.app_api import AppApi
+from clarifai.client.base_auth import BaseAuth
+
+
+class UserApi(BaseAuth):
```
steph-en-m, line 7: I find it a bit repetitive/redundant to emphasize the Api/_api in the naming .e.g. as in `app_api` & `user_api`, should we have them as `apps.py/Apps`, `users.py/Users` so that imports are as `from clarifai.client.apps` or `from clarifai.client.users`?
I missed the design sessions, so it could be that this (user_api, app_api) convention was the chosen one.

clarifai/client/app_api.py
```diff
@@ -0,0 +1,21 @@
+from clarifai_grpc.grpc.api import resources_pb2, service_pb2  # noqa: F401
+
+from clarifai.client.base_auth import BaseAuth
+
+
+class AppApi(BaseAuth):
+  """
+    AppApi is a class that provides access to Clarifai API endpoints related to App information.
+    Inherits from BaseAuth for authentication purposes.
+    """
```
steph-en-m, line 10: Inconsistent indentation for most docstrings throughout the scripts in this PR.

clarifai/client/user_api.py
sainivedh: Naming is not finalised. Now that you say, I agree with removing `api`. But we planned on using respective resource protos, so it should be changed as `user_info` or something

clarifai/client/base_auth.py
sainivedh: By using `BaseAuth`, users are shielded from the complexities of the authhelper. Additionally, it allows for the centralized usage of create_stub across various class interfaces such as inputs, datasets, models, modules, and more

clarifai/client/auth/abc.py
sainivedh: Sure. Will stick to """<>"""

clarifai/client/base_auth.py
steph-en-m: Right, I agree with the centralized usage of `create_stub()`.

On that note, how will users know what keyword args to pass to the class?

sainivedh: The `BaseAuth` is designed to remain hidden from users. Instead, users are encouraged to set the `PAT` (Personal Access Token) as an environment variable at the beginning. The `kwargs` here internally pass `user_id` and `app_id` to obtain the `userDataObject` in other classes.

clarifai/client/user.py
```diff
@@ -0,0 +1,48 @@
...
+  def __init__(self, user_id: str, **kwargs):
+    """
+    Initializes a User object.
+    Args:
+        user_id (str): The user ID for the user to interact with.
+        **kwargs: Additional keyword arguments to be passed to the ClarifaiAuthHelper.
+    """
+    self.kwargs = {**kwargs, 'id': user_id}
+    self.user_info = resources_pb2.User(**self.kwargs)
+    super().__init__(user_id=self.id)
```
steph-en-m, line 22: `self.id` isn't defined anywhere.

I also see `self.id` called in other places.
`id` is quite generic a variable name for use in this case since dataset_id, user_id, app_id, etc can all be ids

steph-en-m (commented):
Some more comments

clarifai/client/user.py
sainivedh: No, The respective protos expects as just `id`
https://github.com/Clarifai/clarifai/blob/8b1af67bbe3d83f06aa799a3b1992a9883b32189/proto/clarifai/api/resources.proto#L2571.

For `App` class `id` refers to app_id and we use `user_id` 
https://github.com/Clarifai/clarifai/blob/8b1af67bbe3d83f06aa799a3b1992a9883b32189/proto/clarifai/api/resources.proto#L75

clarifai/client/user.py
```diff
@@ -0,0 +1,48 @@
...
+    Returns an App object for the specified app ID.
+    Args:
+        app_id (str): The app ID for the app to interact with.
+        **kwargs: Additional keyword arguments to be passed to the App.
+    """
+    kwargs['user_id'] = self.id
+    return App(app_id=app_id, **kwargs)
+
+  def __getattr__(self, name):
+    return getattr(self.user_info, name)
```
sainivedh, line 40: This helps us to get `self.id` and other user_info related params

clarifai/client/user.py
steph-en-m: right, but self.id should still throw an error given it has no definition

sainivedh: We use [getattr](https://github.com/Clarifai/clarifai-python/blob/96b59b56a8e28b24f6a587f3d192a3c06b040120/clarifai/client/user.py#L39) function to get all user_info params

steph-en-m: Cool, hadn't seen that method.

clarifai/client/app.py
```diff
@@ -0,0 +1,21 @@
...
+  def __init__(self, app_id: str, **kwargs):
+    """
+    Initializes a App object.
+    Args:
+        app_id (str): The App ID for the App to interact with.
+        **kwargs: Additional keyword arguments to be passed to the ClarifaiAuthHelper.
+    """
+    self.kwargs = {**kwargs, 'id': app_id}
+    self.app_info = resources_pb2.App(**self.kwargs)
+    super().__init__(app_id=self.id)
```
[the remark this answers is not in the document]
sainivedh: The `getattr` method should be added, but as `App` class dev is in a different ticket left it for now

steph-en-m (approved):
LGTM

PR #137, opened 2023-08-01
[Devx: 177] Listing interface

sainivedh:
## What
- Centralised listing interface for all apps, inputs, datasets, models, workflows, modules ..etc
- Changed `base_auth` to a more appropriate `base_client` 
- Added support for dynamic logging levels


clarifai/client/lister.py
```diff
@@ -0,0 +1,88 @@
...
+    except ValueError:
+      datetime_obj = datetime.strptime(date_str, '%Y-%m-%dT%H:%M:%SZ')
+
+    # Convert the datetime object to a Timestamp object
+    timestamp_obj = Timestamp()
+    timestamp_obj.FromDatetime(datetime_obj)
+
+    return timestamp_obj
+
+  def convert_keys_to_snake_case(self, old_dict: Dict[str, Any],
```
steph-en-m, line 35: Why is this needed?
I would think the integrity of the data fetched from the APIs must be maintained.

clarifai/client/app.py
```diff
@@ -1,21 +1,31 @@
 from clarifai_grpc.grpc.api import resources_pb2, service_pb2  # noqa: F401
 
-from clarifai.client.base_auth import BaseAuth
+from clarifai.client.base_client import BaseClient
```
steph-en-m, line 3: should `base_client.py` just be `base` since it lives under client already?

clarifai/client/lister.py
sainivedh: This is needed for initialising/creating objects from respective resource protos.

The MultiOutputResponse returns response with keys as `userId`, `defaultLanguage` etc... but the protos expects all keys to be snake cased user_id, default_language

clarifai/client/lister.py
```diff
@@ -0,0 +1,88 @@
...
+    Yields:
+        dict: The next item in the listing.
+    """
+    page = 1
+    while True:
+      request_data['page'] = page
+      response = self._grpc_request(endpoint, proto_message(**request_data))
+      dict_response = MessageToDict(response)
+      if response.status.code != status_code_pb2.SUCCESS:
+        raise Exception(f"Listing failed with response {response!r}")
```
steph-en-m, line 81: what's `{response!r}` supposed to represent?

clarifai/client/lister.py
steph-en-m: It looks like you do the convertion after retrieving data from the `List` endpoints

sainivedh: Yes, we get all responses from `List` endpoints and return as respective class object for each of the response

sainivedh: Yes. Only the keys are modified, not the values

steph-en-m: Got it

clarifai/client/lister.py
sainivedh: The !r formatting specifier is used to represent the response object using its repr() method.

steph-en-m (approved):
LGTM

PR #140, opened 2023-08-08
[DEVX:123] Support Dataset Upload

sainivedh:
## What
 - Add Dataset upload functionality to SDK
 - Reuse the current python utils data upload approach
 
## Changes
 - Dataloader is changed to run time proto extraction instead of in-memory 
 - Added a base `ClarifaiDataLoader` class

## Tests
 - Manual data upload test for

      1. COCO det, seg
      2. xView det
      3. Cifar10, Food101 Image cls
      4. IMDB reviews text cls

### DIFF
--- a/clarifai/client/dataset.py
+++ b/clarifai/client/dataset.py
@@ -1,8 +1,24 @@
-from clarifai_grpc.grpc.api import resources_pb2, service_pb2  # noqa: F401
+import time
+import uuid
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from multiprocessing import cpu_count
+from typing import List, Tuple, Union
+
+from clarifai_grpc.grpc.api import resources_pb2, service_pb2
+from clarifai_grpc.grpc.api.status import status_code_pb2, status_pb2
+from google.protobuf.json_format import MessageToDict
+from tqdm import tqdm
+
 from clarifai.client.base import BaseClient
+from clarifai.client.lister import Lister
+from clarifai.datasets.upload.image import (VisualClassificationDataset, VisualDetectionDataset,
+                                            VisualSegmentationDataset)
+from clarifai.datasets.upload.text import TextClassificationDataset
+from clarifai.datasets.upload.utils import load_dataloader, load_zoo_dataloader
+from clarifai.utils.misc import BackoffIterator, Chunker
 
 
-class Dataset(BaseClient):
+class Dataset(Lister, BaseClient):
   """
   Dataset is a class that provides access to Clarifai API endpoints related to Dataset information.
   Inherits from BaseClient for authentication purposes.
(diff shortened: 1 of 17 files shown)


sanjaychelliah (commented):
LGTM! Just minor comments..

clarifai/client/dataset.py
```diff
@@ -16,7 +32,232 @@ def __init__(self, dataset_id: str, **kwargs):
...
+      retry_upload: failed annot upload
+    """
+    retry_upload = []  # those that fail to upload are stored for retries
+    response = self._grpc_request(
+        self.STUB.PostAnnotations,
+        service_pb2.PostAnnotationsRequest(user_app_id=self.user_app_id, annotations=batch_annot),
+    )
+    if response.status.code != status_code_pb2.SUCCESS:
+      try:
+        print(f"Post annotations failed, status: {response.annotations[0].status.details}")
```
sanjaychelliah, line 74: Can this be `logger.info` instead of `print` to maintain consistency throughout the SDK.

clarifai/datasets/upload/zoo/README.md
```diff
@@ -0,0 +1,49 @@
...
+A collection of data preprocessing modules for popular public datasets to allow for compatible upload into Clarifai user app datasets.
+
+## Usage
+
+If a dataset module exists in the zoo, uploading the specific dataset can be easily done by simply creating a python script (or via commandline) and specifying the dataset module name in the `zoo_dataset` parameter of the  `dataset` class, `upload_dataset` method .i.e.
+
+```python
+from clarifai.client.app import App
+
+app = App(user_id="", user_id="")
```
sanjaychelliah, line 12: I think this should be `app_id`.

steph-en-m (commented):
Looking good.

Is the upload sub directory under datasets necessary? (`clarifai.datasets.upload`).
If not , I think we can reduce the nesting and move the files to under datasets.

clarifai/client/dataset.py
```diff
@@ -16,7 +32,232 @@ def __init__(self, dataset_id: str, **kwargs):
...
+      split: split type(train, test, val)
+      module_dir: path to the module directory
+      zoo_dataset: name of the zoo dataset
+    """
+    self.chunk_size = min(128, chunk_size)  # limit max protos in a req
+    self.num_workers: int = min(10, cpu_count())  #15 req/sec rate limit
+    self.annot_num_workers = 4
+    self.max_retires = 10
+    self.task = task
+    datagen_object = None
```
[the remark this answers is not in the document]
sainivedh: > Is the upload sub directory under datasets necessary? (`clarifai.datasets.upload`). If not , I think we can reduce the nesting and move the files to under datasets.

There will be additional functionalities in datasets (export..etc). So had Upload as a sub directory.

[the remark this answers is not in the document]
steph-en-m: export can be a method under this same Dataset class I guess.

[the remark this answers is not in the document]
sainivedh: Cool. For now will define all as attributes under constructor

clarifai/client/dataset.py
```diff
@@ -162,56 +172,69 @@ def _delete_failed_inputs(self, batch_input_ids: List[int]) -> Tuple[List[int],
...
     if retry_annot_protos:
-      chunked_annotation_protos = Chunker(retry_annot_protos, self.chunk_size).chunk()
+      chunked_annotation_protos = Chunker(retry_annot_protos, chunk_size).chunk()
       _ = self._concurrent_annot_upload(chunked_annotation_protos)
 
-  def _data_upload(self) -> None:
-    """Uploads inputs and annotations to clarifai platform dataset."""
-    input_ids = list(range(len(self.dataset_obj)))
-    chunk_input_ids = Chunker(input_ids, self.chunk_size).chunk()
+  def _data_upload(self, task: str, chunk_size: int, dataset_obj: ClarifaiDatasetType) -> None:
```
steph-en-m, line 218: I guess `task` and `chunksize` should remain as part of the constructor because they are clearly used by most methods. Manually redefining them for each method seems redundant.

steph-en-m: I see they are method params for all methods except `delete_failed_inputs()`. They can be defined once and accessed via `self` to reduce redundancy.

ackizilkale (approved):
LGTM as long as the outstanding conversations are resolved.

clarifai/datasets/upload/zoo/README.md
```diff
@@ -0,0 +1,49 @@
+## Datasets Zoo
```
zeiler, line 1: is this really a "zoo" or should we call this "loaders"? zoo/hub is usually a public location where things are shared whereas this folder seems to be more about loading different formats of datasets.

zeiler:
only the one comment on changing this notion of zoo to more of what it seems that folder is which is just different types of loaders of datasets. I didn't do a thorough code review as there are many eyes on it overall.

clarifai/datasets/upload/zoo/README.md
sainivedh: Sure. I think we can rename it to "Dataset Loaders" as its more appropriate. Will change `zoo_dataset` to `dataset_loader`

steph-en-m (approved):
LGTM 👍

PR #143, opened 2023-08-17
[Devx:119] Models Interface

sainivedh:
## What
 - Added functionality to interface with Models(predict, list, versions)
 
 ## How
### Model Predict
```python
from clarifai.client.model import Model

# Model Predict
model = Model(user_id="user_id", app_id="app_id", model_id="model_id")
model_prediction = model.predict_by_url(url="url", input_type="image") # Supports image, text, audio, video

# Customizing Model Inference Output
model = Model(user_id="user_id", app_id="app_id", model_id="model_id",
                  output_config={"min_value": 0.98}) # Return predictions having prediction confidence > 0.98
model_prediction = model.predict_by_filepath(filepath="local_filepath", input_type="text") # Supports image, text, audio, video

model = Model(user_id="user_id", app_id="app_id", model_id="model_id",
                    output_config={"sample_ms": 2000}) # Return predictions for specified interval
model_prediction = model.predict_by_url(url=BEER_VIDEO_URL, input_type="video")
```
### Models Listing
```python
# List all model versions
all_model_versions = model.list_versions()

# Go to specific model version
model_v1 = client.app("app_id").model("model_id", model_version_id="model_version_id")

# List all models in an app
all_models = app.list_models()

# List all models in community filtered by model_type, description
all_llm_community_models = App().list_models(filter_by={"query": "LLM",
                                                        "model_type_id": "text-to-text"}, only_in_app=False)
```

## Tests

[Unit tests](https://github.com/Clarifai/clarifai-python/blob/5052e3bde6bbc5196d06ce39e282af903827ffdd/tests/test_model_predict.py) added for Model predict with different scenarios
 - Predict by filepath, url, bytes
 - Predict by various output configs (max_concepts, min_value, selected_concepts)
 
 ## Note
 Skipping Model unit tests in build as secrets needed for PAT are not yet setup for this repo


README.md
```diff
@@ -34,3 +34,38 @@ dataset = app.create_dataset(dataset_id="demo_dataset")
 # execute data upload to Clarifai app dataset
```
zeiler, line 34: just noticing in the docstring at the top we just say Export your PAT as an environment variable. Let's be sure in all these kind of comments to say CLARIFAI_PAT env variable so we're helping the user at every opportunity where we say those things.

README.md
```diff
@@ -34,3 +34,38 @@ dataset = app.create_dataset(dataset_id="demo_dataset")
...
+                    output_config={"sample_ms": 2000}) # Return predictions for specified interval
+model_prediction = model.predict_by_url(url=BEER_VIDEO_URL, input_type="video")
+```
+### Models Listing
+```python
+# List all model versions
+all_model_versions = model.list_versions()
+
+# Go to specific model version
+model_v1 = client.app("app_id").model("model_id").version("model_version_id")
```
zeiler, line 63: when we do this kind of chained request, is it making an API call for each function call here?

because the alternative would be Model(user_id, app_id, model_id, version_id) which would of course only use one API call. I wonder if we need to add caching in the client with a TTL so that both end up being as efficient?

README.md
sainivedh: Certainly. Will ensure that the PAT requirement is incorporated wherever feasible.

README.md
sainivedh: Yes, for chained request it makes API call for each function(uses `GetApp/Model/DatasetRequest` endpoint). Will discuss with team on adding TTL going further.

clarifai/client/model.py
```diff
@@ -1,22 +1,195 @@
...
+        filepath (str): The filepath to predict.
+        input_type (str): The type of input. Can be 'image', 'text', 'video' or 'audio.
+
+    Example:
+        >>> from clarifai.client.model import Model
+        >>> model = Model(model_id='model_id', user_id='user_id', app_id='app_id')
+        >>> model_prediction = model.predict_by_filepath('/path/to/image.jpg', 'image')
+        >>> model_prediction = model.predict_by_filepath('/path/to/text.txt', 'text')
+    """
+    if input_type not in ['image', 'text', 'video', 'audio']:
```
stmugisha, line 91: It's faster to do set lookups than list lookups. No change required but something you may consider in the future

clarifai/client/model.py
```diff
@@ -1,22 +1,195 @@
...
+      raise UserError('Invalid input type it should be image, text, video or audio.')
+    if not os.path.isfile(filepath):
+      raise UserError('Invalid filepath.')
+
+    with open(filepath, "rb") as f:
+      file_bytes = f.read()
+
+    return self.predict_by_bytes(file_bytes, input_type)
+
+  def predict_by_bytes(self, file_bytes: Union[bytes, str], input_type: str):
```
stmugisha, line 101: shouldn't `file_bytes: Union[bytes, str]` be exclusively bytes rather than having a provision for `str`

stmugisha (commented):
Looking good.
just a few comments

clarifai/client/model.py
```diff
@@ -1,22 +1,195 @@
...
+        >>> model_prediction = model.predict_by_url('url', 'image')
+    """
+    if input_type not in ['image', 'text', 'video', 'audio']:
+      raise UserError('Invalid input type it should be image, text, video or audio.')
+    # TODO will be obtain proto from input class
+    if input_type == "image":
+      input_proto = resources_pb2.Input(
+          data=resources_pb2.Data(image=resources_pb2.Image(url=url)))
+    elif input_type == "text":
+      input_proto = resources_pb2.Input(data=resources_pb2.Data(text=resources_pb2.Text(raw=url)))
```
sanjaychelliah, line 143: This should be `resources_pb2.Text(url=url)`??

clarifai/client/model.py
sainivedh: This is added as union as the `Text` resource proto takes input in both bytes and string.

README.md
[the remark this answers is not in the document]
sainivedh: Not sure I follow you but this helps user to chose a specific `model_version` of `model_id`

README.md
```diff
@@ -34,3 +34,38 @@ dataset = app.create_dataset(dataset_id="demo_dataset")
...
+model_prediction = model.predict_by_url(url="url", input_type="image") # Supports image, text, audio, video
+
+# Customizing Model Inference Output
+model = Model(user_id="user_id", app_id="app_id", model_id="model_id",
+                  output_config={"min_value": 0.98}) # Return predictions having prediction confidence > 0.98
+model_prediction = model.predict_by_filepath(filepath="local_filepath", input_type="text") # Supports image, text, audio, video
+
+model = Model(user_id="user_id", app_id="app_id", model_id="model_id",
+                    output_config={"sample_ms": 2000}) # Return predictions for specified interval
+model_prediction = model.predict_by_url(url=BEER_VIDEO_URL, input_type="video")
```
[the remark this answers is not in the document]
sainivedh: Yes, missed it. Will make it generic

clarifai/client/model.py
sainivedh: Missed it. Will change to url.

README.md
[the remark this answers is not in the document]
stmugisha: When you create a new model object via the API, the model_id and model_version are defined in one call so likewise I think it's best to have the version_id passed as an arg in the model `e.g. (.model(id="model_id", version_id="model_version_id"))` object.

clarifai/client/model.py
stmugisha: since the function is `predict_by_bytes`, there's no need to allow text input.

README.md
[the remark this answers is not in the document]
sainivedh: Oh gotcha, since model_id and model_version_id are not independent we should not separate it out. Will remove `version` function.

stmugisha (approved):
LGTM

PR #149, opened 2023-08-23
[EAGLE-3302] Allow users specify image sizes for image models

stmugisha:
* Refactor triton config types to match triton serializable types. The existing data type specification using strings started throwing data type mismatch errors at times during the triton model repo generation step. This change helps prevent that.
* Allow users specify custom height and width dims. (defaults to [-1, -1, 3])


clarifai/models/model_serving/cli/repository.py
```diff
@@ -40,17 +47,33 @@ def model_upload_init():
...
   parser.add_argument(
       "--repo_dir",
       type=str,
-      default=".",  #curdir
+      default=".",
       required=True,
       help="Directory to create triton repository.")
 
   args = parser.parse_args()
+  if args.image_shape[0] or args.image_shape[1] > 1024:
```
HarmitMinhas96, line 66: Wouldn't this always fail if someone provides a shape?
`args.image_shape[0]` would evaluate to True if it's not 0

clarifai/models/model_serving/model_config/triton_config.py
```diff
@@ -136,60 +155,64 @@ def __post_init__(self):
     Set supported input dims and data_types for
     a given model_type.
     """
-    image_input = InputConfig(name="image", data_type="TYPE_UINT8", dims=[-1, -1, 3])
-    text_input = InputConfig(name="text", data_type="TYPE_STRING", dims=[1])
+    image_dims = self.image_shape
+    image_dims.append(3)  # add channel dim
```
HarmitMinhas96, line 159: Should this assert image_shape is length of 2 first?

clarifai/models/model_serving/model_config/triton_config.py
```diff
@@ -136,60 +155,64 @@ def __post_init__(self):
...
+    image_input = InputConfig(name="image", data_type=DType.TYPE_UINT8, dims=image_dims)
+    text_input = InputConfig(name="text", data_type=DType.TYPE_STRING, dims=[1])
+    # del image_shape as it's a temporary config that's not used by triton
+    del self.image_shape
```
HarmitMinhas96, line 163: I'm confused why this needs to be deleted?

clarifai/models/model_serving/cli/repository.py
stmugisha: It evaluates to the same result as `if args.image_shape[0] > 1024 or args.image_shape[1] > 1024:`
going to update it though since it's not a proper boolean expression.

clarifai/models/model_serving/model_config/triton_config.py
stmugisha: because it's not a valid triton config: just using it to capture shapes to update the input dims which are later serialized.

clarifai/models/model_serving/docs/custom_config.md
```diff
@@ -0,0 +1,33 @@
+## Custom Triton Configurations
+
+The commandline triton model repository generation utils do work with default values for the various triton configurations but a few of these config values can be modified to suit different task specific needs.
+
+* For vision models for instance, different input shapes for the `Height (H)` and `Width (W)` are supported and can be set via the commandline too.i.e.
+```console
+$ clarifai-model-upload-init --model_name <Your model name> \
+		--model_type <select model type from available ones> \
+		--image_shape "H, W"
```
phatvo9, line 9: --image_shape should be `[H, W]` instead `H,W`

stmugisha: Both `[H, W]` and `H, W` are accepted by the `dims_type` (defined under repository.py)

ackizilkale (approved):
Looks good. Please make 1024 a variable

PR #157, opened 2023-08-24
[MRK-3095]Release to Pypi by pushing tags

YvetteZ2017:
Automate the release to Pypi so that a new version can be released when a new tag is pushed.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install .github/workflows/requirements.txt
+          pip install -r .github/workflows/requirements.txt
       - name: Build and publish
         env:
           TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}


.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+on:
+  push:
+    tags:
+      - '[0-9]+.[0-9]+.[0-9a-zA-Z]+'  # Matches 1.2.3, 1.2.3alpha1 etc.
+
+jobs:
+  publish-pypi:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
travertischio, line 15: this should be v3

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v1
```
travertischio, line 17: this should be v4

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+        with:
+          python-version: '3.x'
```
travertischio, line 19: this should probably include the minor version

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel twine
```
travertischio, line 23: we should pin these

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        run: |
+          python setup.py sdist bdist_wheel
+          twine upload --non-interactive dist/*
+  publish-github-release:
+    needs: publish-pypi
+    name: Create Release
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
travertischio, line 36: v3

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Create Release
+        id: create_release
```
travertischio, line 38: this can be removed

.github/workflows/publish.yml
```diff
@@ -0,0 +1,46 @@
...
+        uses: actions/create-release@latest
```
travertischio, line 39: this action is deprecated and we should be using one of the actions that built on this baseline

.github/workflows/publish.yml
```diff
@@ -0,0 +1,42 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install .github/workflows/requirements.txt
```
travertischio, line 23: this is missing an `-r`

travertischio (approved):
the secrets have been added

PR #165, opened 2023-08-30
[EAGLE-3230] [EAGLE-3452] Create tests users can run for triton model upload and fix error in python 3.11

phatvo9:
## What
- Create tests users can run for triton model upload.
- Fix error in python 3.11
## How
- Adding default tests of validating output shape. type and model config. Also allow users to implement their custom tests or even checking model output locally.
- Use default_factory for default value of dataclass


clarifai/models/model_serving/models/default_test.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    return inspect.unwrap(self.triton_model.inference_obj.get_predictions)(
+        self.triton_model.inference_obj, input_data)
+
+  def _get_preprocess(self):
+    """ preprocess if input is image """
+    if "image" in self.triton_model_input_name:
+      h, w, _ = self.triton_model_config.input[0].dims
+      if h > -1 and w > -1:
+        import cv2
```
ackizilkale, line 47: any chance we can use Pillow instead of cv2? cv2 is a headache to install sometimes; I would try to avoid it if possible.

clarifai/models/model_serving/models/default_test.py
```diff
@@ -0,0 +1,227 @@
...
+    np.zeros((100, 100, 3), dtype='uint8'),  #black
+    np.ones((100, 100, 3), dtype='uint8') * 255,  #white
+    np.random.uniform(0, 255, (100, 100, 3)).astype('uint8')  #noise
+]
+
+
+class DefaultTestInferenceModel(unittest.TestCase):
+  """
+  This file contains 2 test cases of:
+  ...
```
ackizilkale, line 25: did you intentionally write ... here?

clarifai/models/model_serving/models/default_test.py
```diff
@@ -0,0 +1,227 @@
...
+        self.assertEqual(
+            type(output), ClassifierOutput,
+            f"Output type must be `ClassifierOutput`, but got {type(output)}")
+        self.assertTrue(
+            _is_valid_logit(output.predicted_scores), "`predicted_scores` must be in range [0, 1]")
+        if self.labels:
+          self.assertEqual(
+              len(output.predicted_scores),
+              len(self.labels),
+              f"`predicted_labels` must equal to {len(self.labels)}, howerver got {len(output.predicted_scores)}"
```
ackizilkale, line 176: spelling for `however`

clarifai/models/model_serving/models/default_test.py
```diff
@@ -0,0 +1,227 @@
...
```
ackizilkale, line 189: spelling for `however`

clarifai/models/model_serving/models/test.py
```diff
@@ -0,0 +1,56 @@
...
+    #process scores to get class id and its score
+    logger.info(result)
+  """
+
+  ########### Initialization. Do not change ###########
+  __test__ = True
+
+  def setUp(self) -> None:
+    logging.info("Initializing...")
+    model_type = "clarifai-model-type"  # your model type
```
ackizilkale, line 48: I wonder if there is a chance to make it obvious that the user needs to edit model_type here

clarifai/models/model_serving/models/default_test.py
phatvo9: I see `opencv-python` already listed in requirements.txt of the package, it would be installed together with `clarifai` and input is always `np.ndarray` so that I prefer opencv

clarifai/models/model_serving/models/test.py
phatvo9: the `model_type` will be added in `test.py` automatically when building the triton repository

clarifai/models/model_serving/models/default_test.py
phatvo9: I intended to write a short description here. I updated the content.

ackizilkale (approved):
LGTM!

clarifai/models/model_serving/README.md
```diff
@@ -14,19 +14,26 @@ $ clarifai-model-upload-init --model_name <Your model name> \
 ```
 2. Edit the `requirements.txt` file with dependencies needed to run inference on your model and the `labels.txt` (if available in dir) with the labels your model is to predict.
 3. Add your model loading and inference code inside `inference.py` script of the generated model repository under the `setup()` and `predict()` functions respectively. Refer to  The [Inference Script section]() for a description of this file.
-4. Generate a zip of your triton model for deployment via commandline.
+4. Testing your implemtation locally by running `<your_triton_folder>/1/test.py` with basic predefined tests.
```
HarmitMinhas96, line 17: *implementation

clarifai/models/model_serving/README.md
```diff
@@ -14,19 +14,26 @@ $ clarifai-model-upload-init --model_name <Your model name> \
...
+To avoid missing dependencies when deploying, recommend to use conda to create clean environment. Then install everything in `requirements.txt`. Follow instruction inside [test.py](./models/test.py) for implementing custom tests. Running the test by using pytest:
```
HarmitMinhas96, line 18: *"Run the test by using pytest:"

HarmitMinhas96 (commented):
Some minor comments, looks good

clarifai/models/model_serving/models/default_test.py
```diff
@@ -0,0 +1,230 @@
...
+
+      field = dataclasses.fields(output)[0].name
+      self.assertEqual(
+          len(self.triton_model_config.output[0].dims),
+          len(getattr(output, field).shape),
+          "Length of 'dims' of config and output must be matched, but get "
+          f"Config {len(self.triton_model_config.output[0].dims)} != Output {len(getattr(output, field).shape)}"
+      )
+
+      if self.model_type == "visual-detector":
```
HarmitMinhas96, line 154: One possibility for these cases is to call predict and check outputs separately. E.g.: visual-classifier and text-classifier can use the same output assertions. text-classifier and text-to-text can use the same input formatting.

HarmitMinhas96: Just a suggestion, not necessary to do now

clarifai/models/model_serving/models/default_test.py
```diff
@@ -0,0 +1,230 @@
...
+                    f"{output_name.name} not in {default_output_names}")
+
+  def test_having_labels(self):
+    if self.model_type in self._required_label_model_types:
+      self.assertTrue(
+          len(self.labels),
+          f"`labels.txt` is empty!. Model type `{self.model_type}` requires input labels in `labels.txt`"
+      )
+
+  def test_inference_with_predefined_inputs(self):
```
deigen, line 119: This is a good test, but I think it may be difficult for the user to debug, for two reasons:
* it's in a library class, not the test file visible to the user, so when it fails they may need to look at the package code to debug
* it's cased with lots of input types other than the one that user is using, making most of the code irrelevant to their model

Another option could be to create a different test class with this test function for each model type, in separate files.  Then when generating the test repo, copy the class for the model type into the test module, so the test code is easily visible and editable (may or may not want to do this second part depending on what it looks like).

Not needed to merge, we could change that later if needed.

phatvo9: Thank you for your suggestions. I believe users can follow the messages in the test and fix their inference code because it has messages for individual assertions.

clarifai/models/model_serving/models/default_test.py
phatvo9: Thank you for your suggestion.

clarifai/models/model_serving/README.md
phatvo9: Updated readme with building env instruction

HarmitMinhas96 (approved):

PR #166, opened 2023-09-05
[DEVX-197]: AddRunners Interface

sainivedh:


clarifai/client/app.py
```diff
@@ -182,6 +183,31 @@ def list_installed_module_versions(self, filter_by: Dict[str, Any] = {}) -> List
...
+
+    Args:
+        filter_by (dict): A dictionary of filters to apply to the list of runners.
+
+    Returns:
+        List[Runner]: A list of Runner objects for the runners in the app.
+
+    Example:
+        >>> from clarifai.client.app import App
+        >>> app = App(app_id="app_id", user_id="user_id")
```
zeiler, line 197: runners are actually not owned apps like other resources, they are owned by the user/org (like a sibling to apps). This is how github does it and in the future we can add runner groups like github has to limits runners to certain apps if needed. So for the API client the runners should be off the User object instead of App.

clarifai/runners/example.py
```diff
@@ -0,0 +1,34 @@
...
+    if data.image.url != "":
+      output.data.text.raw = data.image.url.replace("samples.clarifai.com", "newdomain.com")
+    return output
+
+
+if __name__ == '__main__':
+  # Make sure you set these env vars before running the example.
+  # CLARIFAI_PAT
+  # CLARIFAI_USER_ID
+  # CLARIFAI_APP_ID
```
zeiler, line 31: technically i think this should work without CLARIFAI_APP_ID since the runner is in the user account.

clarifai/client/runner.py
```diff
@@ -0,0 +1,165 @@
...
+  """Base class for remote inference runners. This should be subclassed with the run_input method
+  implemented to process each input in the request.
+
+  Then on the subclass call start() to start the run loop.
+  """
+
+  def __init__(self,
+               runner_id: str,
+               user_id: str = "",
+               app_id: str = "",
```
zeiler, line 36: can remove app_id

zeiler (changes requested):
Thanks for creating this! Added comments about how runners are tied to the user object and not app object.

clarifai/client/app.py
sainivedh: Got it. Now changed to User object instead of App.

zeiler (approved):

PR #167, opened 2023-09-05
[DEVX-198]: Fix Dataset import cache bug

sainivedh:
## Why
 - Added dynamic import of local dir dataset class, so that users doesn't have to reintialise the class eachtime
 
 ```python
 from clarifai.client.dataset import Dataset
d.upload_dataset("visual_clf", "train", "module_dir_path")
d.upload_dataset("visual_detection", "train", "module_dir_path")
 ```
 
image

### DIFF
--- a/clarifai/datasets/upload/utils.py
+++ b/clarifai/datasets/upload/utils.py
@@ -1,10 +1,9 @@
 import importlib
 import inspect
 import os
-import sys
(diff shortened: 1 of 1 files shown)


sanjaychelliah (approved):

ackizilkale (approved):
LGTM

PR #171, opened 2023-09-11
Draft [EAGLE-3454] - Organize model types

phatvo9:
## Why
model types and their config are called in many files but not put in any constant variables

## What
put configs in yaml files


clarifai/models/model_serving/model_config/config.py
```diff
@@ -0,0 +1,300 @@
...
+
+  ### Example:
+  >>> cfg = get_model_config(ModelTypes.text_classifier)
+  >>> custom_triton_config = cfg.make_triton_model_config(**kwargs)
+
+
+  """
+  import os
+  types = ModelTypes()
+  types = types.all
```
HarmitMinhas96, line 290: It seems this is already available in the global var `MODEL_TYPES` from below?

HarmitMinhas96 (approved):

clarifai/models/model_serving/model_config/model_types_config/visual-segmenter.yaml
```diff
@@ -0,0 +1,19 @@
+type: visual-embedder
```
HarmitMinhas96, line 1: This looks like it needs to be updated

clarifai/models/model_serving/models/output.py
```diff
@@ -72,8 +72,8 @@ def __post_init__(self):
     """
     Validate input upon initialization.
     """
-    assert self.predicted_text.ndim == 1, \
-      f"All predictions must be 1-dimensional, Got text-dims: {self.predicted_text.ndim} instead."
+    assert self.predicted_text.ndim == 0, \
+      f"All predictions must be 0-dimensional, Got text-dims: {self.predicted_text.ndim} instead."
```
HarmitMinhas96, line 76: This seems inconsistent with the text-to-text definition that says outputs dims are [1] ? Or I may be understanding this assertion wrong

ackizilkale (approved):
Looks good. Don't forget to address Harmit's unresolved comments above.

clarifai/models/model_serving/models/output.py
phatvo9: It must be 0 otherwise will get error of `EAGLE-3447`

PR #173, opened 2023-09-13
Search V1

sainivedh:
## What
 - Add Search interface in SDK
 - Enable Vector Search (Rank) and also Filter by Parameters
 - Users can rank or Filter by
         -  Concepts
         - Image Bytes
         - Image URL
         - Text 
         - Metadata
         - Geo Points (Long, Lat with radius limit)
 - Supports AND and OR conditions
 - Added validation checks for Schema

## How
```python

from clarifai.client.search import Search
search = Search(user_id='user_id', app_id='app_id', top_k=10, metric='cosine')
```
Alternatively, get the search client via an app:
```python
from clarifai.client.user import User
search = User(user_id="user_id").app(app_id="app_id").search(top_k=10, metric='cosine)
```
Example Usage:
```python
# Vector Search over given image url 
# and filter by images having concepts either of (deer, dog) and , metadata of split as train

query = search.query(ranks=[{"image_url": "https://samples.clarifai.com/metro-north.jpg"}],
                        filters=[{"concepts": [{"name": "deer", "value":1},
                                              {"name": "dog", "value":1}]},
                                {"metadata": {"split": "train"}}], top_k=3)

```

## Tests
 - unit test for filter search and rank search
 - Multiple unit tests for Schema invalid errors


chungisaac (commented):
To add please:
- test case for each key
- test case for AND and OR
- implement filter for input types

isaac-chung:
defer implementing filter for input types to a separate PR.

isaac-chung (approved):
LGTM

PR #179, opened 2023-09-30
Add model predict test case for raw text

isaac-chung:
- Add test case for model prediction for raw text
- Act as extra examples for how to inference on raw text by encoding to UTF-8

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
       - id: isort
   - repo: local
     hooks:
-      - id: autofalke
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -18,7 +18,7 @@ repos:
       - id: isort
   - repo: local
     hooks:
-      - id: autofalke
```
sainivedh, line 21: Oh no... Nice catch!

sainivedh (approved):
Looks Good

PR #180, opened 2023-10-02
update requirements and env handling in model upload

deigen:
In https://github.com/Clarifai/clarifai/pull/16601, we removed the need for users to specify base conda envs.  Instead, we determine which env to use based on the requirements.txt.  This PR is updates the client accordingly:

* removes conda yaml files
* adds torch dependency to requirements.txt
* updates documentation


ackizilkale (approved):
looks good

PR #185, opened 2023-10-05
VERSION_9_9_0

sainivedh:
- Workflow Create Bug fixes
 - Supports Annotation Search


ackizilkale:
looks good for now. But next time we do this we should make sure that clarifai/versions.py and setup.py read from VERSION file. Feels silly to change this in 3 places.

ackizilkale (approved):

sainivedh:
> looks good for now. But next time we do this we should make sure that clarifai/versions.py and setup.py read from VERSION file. Feels silly to change this in 3 places.

Agreed. Will change this on next version PR

PR #186, opened 2023-10-09
[DEVX-208] Duplicate Input-ID Bug

sanjaychelliah:
# what
[Input ID bug](https://clarifai.atlassian.net/browse/DEVX-208?atlOrigin=eyJpIjoiZWI2YmM1YmMyOGIxNDliNGE5OTRiNzNlYmY3MmQ4YWMiLCJwIjoiaiJ9)

# How

- Random id as input ID while uploading from CSV for now.
- Imagenet loader input ID updated

# TODO

- Receiving input ID from the CSV itslef.(When handling multiple columns from CSV)

### DIFF
--- a/clarifai/client/input.py
+++ b/clarifai/client/input.py
@@ -289,7 +289,7 @@ class Inputs(Lister, BaseClient):
                           csv_type: str = 'raw',
                           dataset_id: str = None,
                           labels: str = True) -> List[Text]:
-    """Create input protos from cscv.
+    """Create input protos from csv.
 
     Args:
         csv_path (str): Path to the csv file.
--- a/clarifai/client/input.py
+++ b/clarifai/client/input.py
@@ -317,7 +317,7 @@ class Inputs(Lister, BaseClient):
         else:
           labels = None
 
-        input_id = f"{dataset_id}-{id}"
+        input_id = uuid.uuid4().hex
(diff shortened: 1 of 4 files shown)


sainivedh (changes requested):
I think it must be changed in `image.py` and `text.py`
https://github.com/Clarifai/clarifai-python/blob/master/clarifai/datasets/upload/image.py
https://github.com/Clarifai/clarifai-python/blob/master/clarifai/datasets/upload/text.py

Have you run tests, by uploading same data twice into same app with different datasets?

sanjaychelliah:
> I think it must be changed in `image.py` and `text.py` https://github.com/Clarifai/clarifai-python/blob/master/clarifai/datasets/upload/image.py https://github.com/Clarifai/clarifai-python/blob/master/clarifai/datasets/upload/text.py
> 
> Have you run tests, by uploading same data twice into same app with different dataset ids?

Handled two bugs and tested both..

- CSV input id generation bug.
- Same data uploading in different dataset ID s in the same app..

Made changes for the second bug.. 👍

sainivedh (approved):
lgtm. 
Not relevant to this but a minor spell issue in `inputs.py`
https://github.com/Clarifai/clarifai-python/blob/910de788045c52c718fa42be38df4d18e6ced104/clarifai/client/input.py#L292

PR #193, opened 2023-10-12
Add Ruff to pre-commit hooks

isaac-chung:
**Major change**
- Replace Autoflake with [Ruff](https://github.com/astral-sh/ruff):
> Beyond the defaults, Ruff re-implements some of the most popular Flake8 plugins and related code quality tools, including:
[autoflake](https://pypi.org/project/autoflake/)

- Disabled error E501 (lines too long) as there were too many. Will leave this for later
- Fix all other errors from Ruff

**Side changes**
- Fix some tests revealed by Ruff errors
- Timestamp added to app_id for some tests as it was likely that previous test failures did not clean up the app properly

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,6 +16,11 @@ repos:
     rev: 5.12.0
     hooks:
       - id: isort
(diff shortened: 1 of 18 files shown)


sainivedh (approved):
Looks good.

PR #194, opened 2023-10-12
[DEVX-215]: Lister_update

sanjaychelliah:
# Features

- Added Pagination for all the listers.
```python
all_workflows = list(app.list_workflows(page_no=1, per_page=10))
```

- [Updated yield for all the listers.](https://clarifai.atlassian.net/browse/DEVX-218?atlOrigin=eyJpIjoiMDg0YjRiZTJlNTZmNGY4OWJjZmUyMjJhNjYxYjBhNDciLCJwIjoiaiJ9)
- List Concepts was added.
```python
concepts = list(app.list_concepts())
```

- Updated dataset ID, Input type filters for `list_inputs`
```python
input_object = app.inputs()
image_filterd_inputs = list(input_object.list_inputs(dataset_id = 'xxx', input_type='image'))
```

- [List annotations added](https://clarifai.atlassian.net/browse/DEVX-233?atlOrigin=eyJpIjoiMjZjNGRkZDhkMzU0NDc2MTg3MmFmOTI3MTZiYjc3MjciLCJwIjoiaiJ9)
```python
all_inputs = list(input_obj.list_inputs(input_type='image'))
all_annotations = list(input_obj.list_annotations(batch_input=all_inputs))
```

## Test Coverage

- Lister tests are updated to include pagination, type filter, list concepts and list annotations
- Dataset.upload_dataset() functionality coverage.


README.md
```diff
@@ -63,6 +63,7 @@ client = User(user_id="user_id")
 
 # Get all apps
 apps = client.list_apps()
```
sainivedh, line 65: can we rename it to `apps_generator_obj` or something else

clarifai/client/lister.py
```diff
@@ -12,22 +12,30 @@ class Lister(BaseClient):
   def __init__(self, page_size: int = 16):
     self.default_page_size = page_size
```
sainivedh, line 13: Is this still needed ?

clarifai/client/model.py
```diff
@@ -90,36 +90,43 @@ def create_model_version(self, **kwargs) -> 'Model':
 
     return Model(**kwargs)
 
-  def list_versions(self) -> List['Model']:
+  def list_versions(self, page_no: int = None,
+                    per_page: int = None) -> Generator['Model', None, None]:
```
sainivedh, line 94: This goes for all. What does `None` in `Generator['Model', None, None]` represent here, should we use `Union` here?!

sainivedh (commented):
Looks good. Added some comments

clarifai/client/model.py
sanjaychelliah: [This](https://docs.python.org/3/library/typing.html#:~:text=If%20your%20generator%20will%20only%20yield%20values%2C%20set%20the%20SendType%20and%20ReturnType%20to%20None%3A) is the reason it was kept as None. Since we're only yielding values, send_type and return_type are set to None.

clarifai/client/lister.py
sanjaychelliah: Yes. For scenarios where per_page is not set and the page is set, the user can set it in init itself. And also search class makes use of this..

README.md
sanjaychelliah: Thanks for the catch 👍

sainivedh (approved):
lgtm

PR #201, opened 2023-10-19
[EAGLE-3552][EAGLE-3567] - Add inference params and multimodal-embedder model

phatvo9:
[EAGLE-3552] - Add inference parameters:
* Modify model.py and inference.py to receives them
* Generate JSON file for uploading
* Test with inference parameters
* update doc

[EAGLE-3552] - add multimodal-embedder model type

[EAGLE-3552]: https://clarifai.atlassian.net/browse/EAGLE-3552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EAGLE-3552]: https://clarifai.atlassian.net/browse/EAGLE-3552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/clarifai/models/model_serving/docs/inference_parameters.md
+++ b/clarifai/models/model_serving/docs/inference_parameters.md
@@ -1,21 +1,19 @@
-## Inference paramters
+## Inference paramaters
 
-When making prediction, you may need to change some paramters to adjust the result. Those paramters will be passed through `paramters()` of a request in triton python model.
+When making prediction, you may need to change some paramaters to adjust the result. Those paramaters will be passed through `paramaters()` of a request in triton python model.
 
-In order to send it to `**kwargs` of `get_predictions` in `inference.py`, there are 2 ways:
-1. You can send any arbitrary parameters via clarifai API.
-2. You can define some parameters and they will be visible and adjustable on Clarifai model views.
+In order to send it to `**kwargs` of `get_predictions` in `inference.py`, you can define some parameters and they will be visible and adjustable on Clarifai model view.
 
-This document helps you to create your inference parameters that can be visibale and adjustable easily on Clarifai platform. `Again, you can still send any parameters via API but undefined parameters won't appear on Clarifai UI`. The defined parameters will be sent as `json` file when you use `clarifai-upload-model` cli.
+This document helps you to create your inference parameters that can be visibale and adjustable easily on Clarifai platform. The defined parameters will be sent as `json` file when you use `clarifai-upload-model` cli.
 
 ### JSON file structure:
 The file contains a list of object has 4 fields:
 * `path` (str): name of your parameter, it must be valid as python variable
-* `field_type` (int): the parameter data type is one of {1,2,3}, it means {boolean, string, number} respectively. `Number` means `int` or `float`
+* `field_type` (int): the parameter data type is one of {1,2,21,3}, it means {boolean, string, encrypted_string, number} respectively. `Number` means `int` or `float`. `Encrypted_string` is simple a string can be used to store your secrets like API key...
 * `default_value`: a default value of the parameter.
 * `description` (str): short sentence describes what the parameter does
 
-An example of 3 parameters:
+An example of 4 parameters:
 ```json
 [
   {
@@ -35,17 +33,24 @@ An example of 3 parameters:
     "field_type": 3,
     "default_value": 9.9,
     "description": "a float number variable"
-  }
+  },
+  {
+    "path": "secret_string_var",
+    "field_type": 21,
+    "default_value": "API_KEY",
+    "description": "a string variable contains secret like API key"
+  },
 ]
 ```
 
 ### Generate JSON file
 1. Manually create the file based on above structure
 2. By code:
+
+#### 2.1. Fully setup
 ```python
 from clarifai.models.model_serving.model_config.inference_parameter import InferParamManager, InferParam, InferParamType
 
-# 2.1. Fully setup
 params = [
   InferParam(
     path="boolean_var",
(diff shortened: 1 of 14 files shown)


clarifai/models/model_serving/docs/inference_parameters.md
```diff
@@ -0,0 +1,82 @@
+## Inference paramters
```
HarmitMinhas96, line 1: Few instances in this file:
*paramters -> parameters

similarly in clarifai/models/model_serving/models/test.py

clarifai/models/model_serving/models/model_types.py
```diff
@@ -234,3 +234,32 @@ def parse_predictions(self, input_data: np.ndarray):
     return inference_response
 
   return parse_predictions
+
+
+def multimodal_embedder(func: Callable):
+  """
+  Visual embedder type output parser.
+  Generates embeddings for an input image.
```
HarmitMinhas96, line 242: This should mention this supports both image and/or text

clarifai/models/model_serving/model_config/inference_parameter.py
```diff
@@ -0,0 +1,109 @@
+import json
+from dataclasses import asdict, dataclass, field
+from typing import Any, List
+
+
+@dataclass(frozen=True)
+class InferParamType:
+  BOOL: int = 1
+  STRING: int = 2
+  NUMBER: int = 3
```
HarmitMinhas96, line 10: Very soon we're going to have `ENCRYPTED_STRING = 21;` support which will be needed here. I think it's ok to get ahead and add that now. That would be used for things like API Keys where we never want to return it via the API. The user would provide a string value. I see the code currently depends on their only being one string field type, so maybe we can always default to string, but have a way for users to specify it's an encrypted string too.
cc @wemoveon2

clarifai/models/model_serving/model_config/inference_parameter.py
```diff
@@ -0,0 +1,109 @@
...
+    return list_params
+
+  def export(self, path: str):
+    list_params = self.get_list_params()
+    with open(path, "w") as fp:
+      json.dump(list_params, fp, indent=2)
+
+  def validate(self, **kwargs) -> dict:
+    output_kwargs = {k: v.default_value for k, v in self._dict_params.items()}
+    assert not (kwargs != {} and self.params == []), "kwargs are rejected since `params` is empty"
```
HarmitMinhas96, line 86: Might be simpler written:
```kwargs == {} or self.params != []```

File: clarifai/models/model_serving/docs/inference_parameters.md [hunk 5]
```diff
@@ -0,0 +1,82 @@
...
+
+When making prediction, you may need to change some paramters to adjust the result. Those paramters will be passed through `paramters()` of a request in triton python model.
+
+In order to send it to `**kwargs` of `get_predictions` in `inference.py`, there are 2 ways:
+1. You can send any arbitrary parameters via clarifai API.
```
HarmitMinhas96, line 6: We're removing this, they will need to be defined by the model owner

clarifai/models/model_serving/docs/inference_parameters.md
```diff
@@ -0,0 +1,82 @@
...
+2. You can define some parameters and they will be visible and adjustable on Clarifai model views.
```
HarmitMinhas96, line 7: *visible

HarmitMinhas96 (commented):
Could you possibly split the multi-modal embedder into it's own smaller PR?

clarifai/models/model_serving/model_config/inference_parameter.py
wemoveon2: yup add `ENCRYPTED_STRING` here and i'll make the changes to BE to support the field type.

clarifai/models/model_serving/models/default_test.py
```diff
@@ -30,58 +32,73 @@ class DefaultTestInferenceModel(unittest.TestCase):
...
+    _kwargs = self.inference_parameters.validate(**kwargs)
+    return inspect.unwrap(self.triton_python_model.inference_obj.get_predictions)(
+        self.triton_python_model.inference_obj, input_data, **_kwargs)
 
-  def _get_preprocess(self):
+  def _get_preprocess(self, input):
     """ preprocess if input is image """
-    if "image" in self.triton_model_input_name:
-      h, w, _ = self.triton_model_config.input[0].dims
+    if "name" in input.name:
```
HarmitMinhas96, line 51: I think "name" is supposed to remain "image" here?

clarifai/models/model_serving/docs/inference_parameters.md
```diff
@@ -0,0 +1,134 @@
...
+When making prediction, you may need to change some paramaters to adjust the result. Those paramaters will be passed through `paramaters()` of a request in triton python model.
+
+In order to send it to `**kwargs` of `get_predictions` in `inference.py`, you can define some parameters and they will be visible and adjustable on Clarifai model view.
+
+This document helps you to create your inference parameters that can be visibale and adjustable easily on Clarifai platform. The defined parameters will be sent as `json` file when you use `clarifai-upload-model` cli.
+
+### JSON file structure:
+The file contains a list of object has 4 fields:
+* `path` (str): name of your parameter, it must be valid as python variable
+* `field_type` (int): the parameter data type is one of {1,2,21,3}, it means {boolean, string, encrypted_string, number} respectively. `Number` means `int` or `float`. `Encrypted_string` is simple a string can be used to store your secrets like API key...
```
HarmitMinhas96, line 12: "Encrypted_string is simple a string can be used to store your secrets like API key..."
->
"Encrypted_string is a string that can be used to store your secrets, like API key. The API will not return the values for this as plaintext."

HarmitMinhas96 (approved):

clarifai/models/model_serving/docs/inference_parameters.md
phatvo9: thanks for your suggestion

clarifai/models/model_serving/docs/inference_parameters.md
phatvo9: Sure, error will arise if user uses undefined parameters in test cases

HarmitMinhas96 (approved):

PR #210, opened 2023-11-14
[DEVX-289]: Fix passing BaseURL info

sainivedh:
## What
 -  Fix BaseURL info not passing down to other resources
 - Now users can be able to do
 ```python
user = User(user_id = user_id, base_url = base_url)
app = user.app(app_id = app_id)
model = app.model(model_id = model_id,model_version_id = version_id)
```


sanjaychelliah (approved):
LGTM

PR #212, opened 2023-11-16
change pycocotools to extras require

sainivedh:
Relax pandas, numpy, pyyaml, pillow, rich versions 
change pycocotools to extras_require (`pip install "clarifai[all]"`)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ setuptools.setup(
     python_requires='>=3.8',
     install_requires=install_requires,
     extras_require={
-        'data': ["pycocotools=2.0.6"],
+        'all': ["pycocotools=2.0.6"],
     },
     entry_points={
         "console_scripts": [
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -34,6 +34,9 @@
     license="Apache 2.0",
     python_requires='>=3.8',
     install_requires=install_requires,
+    extras_require={
+        'data': ["pycocotools=2.0.6"],
```
sanjaychelliah, line 38: can we have this as `pip install "clarifai[all]"` or  this is fine `pip install "clarifai[data]"` ?

sainivedh: Both seems fine, since data-utilities are planned next, I thought of having `data`

sainivedh: @sriram-clarifai your thoughts please

sanjaychelliah (approved):
LGTM

PR #217, opened 2023-11-22
[EAGLE-3652] - vLLM example

phatvo9:


clarifai/models/model_serving/examples/vllm/example/1/test.py
```diff
@@ -0,0 +1,64 @@
...
+  3. If your input is `image` and you have set custom size of it when building model repository,
+  call `self.preprocess(image)` to obtain correct resized input
+  4. Run this test by calling
+  ```bash
+  pytest ./your_triton_folder/1/test.py
+  #to see std output
+  pytest --log-cli-level=INFO  -s ./your_triton_folder/1/test.py
+  ```
+
+  ### Examples:
```
ackizilkale, line 24: We need to update these examples if we will show them here. text_to_image and visual_classifier are unlikely helpful.

ackizilkale (changes requested):
Can we update the examples on the test page?

clarifai/models/model_serving/examples/vllm/example/1/test.py
phatvo9: The examples in the test.py was generated automatically when initializing model repository.

ackizilkale (approved):
Looks good!

PR #218, opened 2023-11-22
[DEVX-288] Dataset Upload Status

sainivedh:


clarifai/datasets/upload/utils.py
```diff
@@ -41,3 +45,200 @@ def load_module_dataloader(module_dir: Union[str, os.PathLike], **kwargs) -> Cla
...
+    local_annotations_dict = dict(concepts=0, bboxes=0, polygons=0)
+    for i in range(local_inputs_count):
+      if task == "visual_segmentation":
+        local_annotations_dict["polygons"] += len(self.dataloader[i].polygons)
+      elif task == "visual_detection":
+        local_annotations_dict["bboxes"] += len(self.dataloader[i].bboxes)
+      elif task == "text_classification":
+        local_annotations_dict["concepts"] += len(self.dataloader[i].labels)
+      elif task == "visual_classification" or task == "visual_captioning":
+        local_annotations_dict["concepts"] += len(self.dataloader[i].labels)
```
isaac-chung, line 118: as more tasks and annotation types may be supported in the future, we could define a dict that maps the task -> annot_dict key -> attribute from data_item. We can then refactor this to simply read from the dict and use getattr on the data_item. e.g.
```python
## Put this in constants file somewhere
some_dict = {
    "visual_segmentation": {"polygons": "polygons"},
    "some_task": {"key": "attr"}
}
.....

## This for loop
for ...... :
    key, attr = [(k, v) for k, v in some_dict.get(task).items()][0]
    local_annotations_dict[key] += len(getattr(self.dataloader[i], attr))
```

Something like that. That way we will not have to add an `elif` every time we get a new annotation type or a task type.

isaac-chung (commented):
The display looks nice :)

clarifai/client/dataset.py
```diff
@@ -384,6 +387,61 @@ def upload_from_folder(self,
           folder_path=folder_path, dataset_id=self.id, labels=labels)
     self.input_object._bulk_upload(inputs=input_protos, batch_size=batch_size)
 
+  def get_upload_status(self, dataloader: Type[ClarifaiDataLoader],
+                        delete_version: bool = False) -> None:
+    """Displays the upload status of the dataset.
+
+    Args:
+        dataloader (Type[ClarifaiDataLoader]): ClarifaiDataLoader object
+        delete_version (bool): True if you want to delete the version after getting the upload status
```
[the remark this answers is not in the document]
sainivedh: Since this creates a new dataset version every time. Provided an option so they can keep deleting the version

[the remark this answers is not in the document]
isaac-chung: Oh. I must have missed the upload line. In that case, please update both the function name and docstrings to reflect that. I'll add suggestions.

clarifai/client/dataset.py
```diff
@@ -384,6 +387,61 @@ def upload_from_folder(self,
           folder_path=folder_path, dataset_id=self.id, labels=labels)
     self.input_object._bulk_upload(inputs=input_protos, batch_size=batch_size)
 
+  def get_upload_status(self, dataloader: Type[ClarifaiDataLoader],
```
isaac-chung, line 390:
```suggestion
  def upload_and_get_status(self, dataloader: Type[ClarifaiDataLoader],
```

clarifai/client/dataset.py
```diff
@@ -384,6 +387,61 @@ def upload_from_folder(self,
...
+                        delete_version: bool = False) -> None:
+    """Displays the upload status of the dataset.
```
isaac-chung, line 392:
```suggestion
    """Creates a new dataset version and displays the upload status of the dataset.
```

clarifai/client/dataset.py
sainivedh: `get_upload_status` fits better, this function only just gives statistics of upload process, upload is done via `upload_dataset()`.

isaac-chung: In what order would the user call these functions? Can we include the example usage in the docstring?
From this example it is not clear when a) `upload_dataset()` is called and b) why `get_upload_status()` is called on an empty dataset as it was just instantiated.
```
>>> from clarifai.client.dataset import Dataset
 >>> dataset = Dataset(dataset_id='dataset_id', user_id='user_id', app_id='app_id')
>>> dataset.get_upload_status(dataloader)
```

sainivedh: Also, we could call this feature at the end of [upload_dataset()](https://github.com/Clarifai/clarifai-python/blob/68bcbd8890e19422d530144fa1feb7b48f4f056f/clarifai/client/dataset.py#L313). So by default user sees the stats after upload being done.

clarifai/datasets/upload/utils.py
```diff
@@ -41,3 +45,200 @@ def load_module_dataloader(module_dir: Union[str, os.PathLike], **kwargs) -> Cla
...
+  def get_uploaded_dataset_stats(self) -> Tuple[Dict[str, int], Dict[str, int]]:
+    """Parse the response from the server for the dataset version metrics groups.
+
+    Returns:
+      uploaded_inputs_dict (Dict[str, int]): The input statistics for the dataset.
+      uploaded_annotations_dict (Dict[str, int]): The annotation statistics for the dataset.
+    """
+    dataset_statistics = []
+    uploaded_inputs_dict = {}
+    uploaded_annotations_dict = {"concepts": 0, "bboxes": 0, "polygons": 0}
```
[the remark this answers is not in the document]
sainivedh: Done

clarifai/datasets/upload/utils.py
sainivedh: Added this

clarifai/client/dataset.py
sainivedh: Done

clarifai/client/dataset.py
isaac-chung: Ya that would be great.

clarifai/client/dataset.py
```diff
@@ -384,6 +387,61 @@ def upload_from_folder(self,
...
+        self.delete_version(dataset_version_id)
+        raise Exception("Failed to get dataset metrics {}".format(dataset_metrics_response.status))
+
+      dict_response = MessageToDict(dataset_metrics_response)
+      if len(dict_response.keys()) == 1 and time.time() - start_time < 60 * 10:  # 10 minutes
+        self.logger.info("Crunching the dataset metrics. Please wait...")
+        time.sleep(next(backoff_iterator))
+        continue
+      else:
+        if time.time() - start_time > 60 * 10:  # 10 minutes
```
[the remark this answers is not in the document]
isaac-chung: Thanks, and sorry I wasn't super clear. I was hoping for this parameter to be in the function signature as well, with the same default value.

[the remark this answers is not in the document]
sainivedh: Added `timeout` as arg

isaac-chung (approved):
Looks good! Thanks for iterating.

PR #221, opened 2023-11-23
[DEVX-300] Support PAT as arg

sainivedh:
## What
 - Support `PAT` as parameter while initialising resources in addition to setting as env variable
 - Changed necessary methods in `Inputs` class to staticmethods

## How
```python
model = Model(url="https://clarifai.com/openai/chat-completion/models/GPT-3_5-turbo", pat="CLARIFAI_PAT")


print(model.predict_by_bytes(b"Hello World", "text"))
```

## Tests
 - Initialised respective `User`, `App`, `Model`, `Workflow` classes w.r.t `PAT` in `test_app.py`, `test_model_predict.py` and `workflow/test_predict.py`

### DIFF
--- a/clarifai/client/dataset.py
+++ b/clarifai/client/dataset.py
@@ -34,10 +34,10 @@ class Dataset(Lister, BaseClient):
   """Dataset is a class that provides access to Clarifai API endpoints related to Dataset information."""
 
   def __init__(self,
-               url: str = "",
-               dataset_id: str = "",
+               url: str = None,
+               dataset_id: str = None,
                base_url: str = "https://api.clarifai.com",
-               pat: str = "",
+               pat: str = None,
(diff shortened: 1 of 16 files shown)


clarifai/client/dataset.py
```diff
@@ -37,13 +37,15 @@ def __init__(self,
                url: str = "",
                dataset_id: str = "",
                base_url: str = "https://api.clarifai.com",
+               pat: str = "",
```
isaac-chung, line 40: Perhaps a more pythonic way would be to default these to `None` instead of `""`. That way in checks you can do `if not param` as well, for example.

isaac-chung (commented):
Much cleaner! One discussion comment.

isaac-chung (approved):

PR #225, opened 2023-11-24
Remove pytest requirement

Harmon758:
As far as I can tell, pytest is not a necessary requirement for the library itself.

It's already included in tests/requirements.txt, and in fact, is overridden (from ==7.4.1 in requirements.txt to ==7.1.2 in tests/requirements.txt) in the `Run tests` GitHub Actions workflow (e.g. https://github.com/Clarifai/clarifai-python/actions/runs/6970572930/job/18968766681#step:4:179).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,6 @@ omegaconf==2.2.3
 opencv-python==4.7.0.68
 tritonclient==2.34.0
 rich>=13.4.2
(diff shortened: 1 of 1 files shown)


sainivedh (approved):
Looks good. Thanks for creating this!

sanjaychelliah (approved):

PR #227, opened 2023-11-28
[EAGLE-3698] - model upload handles multiple batch

phatvo9:
### Why
For now model only predicts one by one input even sending a batch with size >1.
### How
- `get_predictions()` method in` inference.py` will take a list of inputs instead of single input.
- Update examples with batch input.
- Update doc
### Other updates:
- insert triton decorator function when initializing model repository, so user won't forget to do this.
- enable infer param description from_kwargs.

### Note:
Models generated by a lower version will not function on this version.


HarmitMinhas96 (commented):
This PR is currently very large. Can it be broken down into more manageable chunks for review?
E.g.:
1. Add multiple batches handling in one PR (maybe two if it can be logicially split)
2. Add new model type example text-embedder
3. Add new model type example multimodal-embedder
4. Add vllm example

phatvo9:
> This PR is currently very large. Can it be broken down into more manageable chunks for review? E.g.:
> 
> 1. Add multiple batches handling in one PR (maybe two if it can be logicially split)
> 2. Add new model type example text-embedder
> 3. Add new model type example multimodal-embedder
> 4. Add vllm example
> 
> Or the model type and vllm examples can be added first if you prefer

Broke down into #236 : update code for batching and update old examples and #237 added text-embedder and multimodal-embedder examples
Since vllm example is merged #217, so I added it together with old examples.

PR #228, opened 2023-11-28
[Bug] Fix root logger handler

### LINKED ISSUE #220: Handler added to root logger
This library adds a `RichHandler` to the root logger through `logging.basicConfig` every time `get_logger` is called (which seems to be every time any common class/object is initialized):
https://github.com/Clarifai/clarifai-python/blob/dbdcb493bdb9255244276086bd193a305e2cbcc0/clarifai/utils/logging.py#L76-L90

There doesn't seem to be any option to configure this, and this can cause conflicts with a user's own logging setup and/or those of other libraries. (Especially since `RichHandler` seems to add a lot of formatting that a user's own logging setup might not support)

Instead, this library should configure and add handlers to its own loggers instead of the root logger.

sainivedh:
## Why
 - Fixes #220 
 
## What
 - Instead of adding `RichHandler` to root logging, now it is being added to SDK specific loggers
 
## Tests
 - Added test_misc.py for the same 
 
image

### DIFF
--- a/clarifai/utils/logging.py
+++ b/clarifai/utils/logging.py
@@ -73,11 +73,21 @@ def _get_library_name() -> str:
   return __name__.split(".")[0]
 
 
-def _configure_logger(logger_level: str = "ERROR") -> None:
-  logging.basicConfig(
-      level=logger_level,
-      datefmt='%Y-%m-%d %H:%M:%S',
-      handlers=[RichHandler(rich_tracebacks=True)])
+def _configure_logger(name: str, logger_level: str = "ERROR") -> None:
+  """Configure the logger with the specified name."""
+
+  logger = logging.getLogger(name)
+  logger.setLevel(logger_level)
+
+  # Remove existing handlers
+  for handler in logger.handlers[:]:
+    logger.removeHandler(handler)
+
+  # Add the new rich handler and formatter
+  handler = RichHandler(rich_tracebacks=True)
+  formatter = logging.Formatter('%(name)s:  %(message)s')
+  handler.setFormatter(formatter)
+  logger.addHandler(handler)
 
 
 def get_logger(logger_level: str = "ERROR", name: Optional[str] = None) -> logging.Logger:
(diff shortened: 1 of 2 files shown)


clarifai/utils/logging.py
```diff
@@ -73,11 +73,21 @@ def _get_library_name() -> str:
   return __name__.split(".")[0]
 
 
-def _configure_logger(logger_level: str = "ERROR") -> None:
-  logging.basicConfig(
-      level=logger_level,
-      datefmt='%Y-%m-%d %H:%M:%S',
```
isaac-chung, line 79: How is the date format set now?

clarifai/utils/logging.py
```diff
@@ -73,11 +73,21 @@ def _get_library_name() -> str:
...
-      handlers=[RichHandler(rich_tracebacks=True)])
+def _configure_logger(name: str, logger_level: str = "ERROR") -> None:
```
isaac-chung, line 76:
```suggestion
def _configure_logger(name: str, logger_level: str = logging.ERROR) -> None:
```

clarifai/utils/logging.py
```diff
@@ -73,11 +73,21 @@ def _get_library_name() -> str:
...
+    logger.removeHandler(handler)
+
+  # Add the new rich handler and formatter
+  handler = RichHandler(rich_tracebacks=True)
+  formatter = logging.Formatter('%(name)s:  %(message)s')
+  handler.setFormatter(formatter)
+  logger.addHandler(handler)
 
 
 def get_logger(logger_level: str = "ERROR", name: Optional[str] = None) -> logging.Logger:
```
isaac-chung, line 93:
```suggestion
def get_logger(logger_level: str = logging.ERROR, name: Optional[str] = None) -> logging.Logger:
```

clarifai/utils/logging.py
sainivedh: Great catch. Added back the same format now.

isaac-chung (approved):

sanjaychelliah (approved):

PR #231, opened 2023-11-29
[DEVX-301]: Removing internal_only Training Params

sanjaychelliah:
# what

- Removing internal only training params.
- Removing internal only training model type.
- Changing pytest test scripts according to it..


sainivedh (approved):
lgtm

PR #236, opened 2023-11-30
[EAGLE-3698]-split-handle batch

phatvo9:
# What
Split part of #227, handle batch

### DIFF
--- a/clarifai/models/model_serving/cli/deploy_cli.py
+++ b/clarifai/models/model_serving/cli/deploy_cli.py
@@ -13,7 +13,7 @@
 """Commandline interface for model upload utils."""
 import argparse
 
-from clarifai.auth.helper import ClarifaiAuthHelper
+from clarifai.client.auth.helper import ClarifaiAuthHelper
 from clarifai.models.api import Models
 from clarifai.models.model_serving.model_config import MODEL_TYPES, get_model_config
 from clarifai.models.model_serving.model_config.inference_parameter import InferParamManager
--- a/clarifai/models/model_serving/models/inference.py
+++ b/clarifai/models/model_serving/models/inference.py
@@ -25,7 +25,7 @@ class InferenceModel:
     #self.model: Callable = <load_your_model_here from checkpoint or folder>
 
   #Add relevant model type decorator to the method below (see docs/model_types for ref.)
-  def get_predictions(self, input_data, **kwargs):
+  def get_predictions(self, input_data: list, **kwargs):
     """
     Main model inference method.
 
--- a/clarifai/models/model_serving/models/inference.py
+++ b/clarifai/models/model_serving/models/inference.py
@@ -10,6 +10,11 @@
 import os
 from pathlib import Path
 
+from clarifai.models.model_serving.model_config import (  # noqa # pylint: disable=unused-import
+    ModelTypes, get_model_config)
+
+#config = get_model_config("clarifai-model-type")
+
 
 class InferenceModel:
   """User model inference class."""
@@ -24,19 +29,22 @@ class InferenceModel:
     #self.checkpoint_path: Path = os.path.join(self.base_path, "your checkpoint filename/path")
     #self.model: Callable = <load_your_model_here from checkpoint or folder>
 
-  #Add relevant model type decorator to the method below (see docs/model_types for ref.)
-  def get_predictions(self, input_data: list, **kwargs):
+  #@config.inference.wrap_func
+  def get_predictions(self, input_data: list, **kwargs) -> list:
     """
     Main model inference method.
 
     Args:
     -----
-      input_data: A single input data item to predict on.
+      input_data: A list of input data item to predict on.
         Input data can be an image or text, etc depending on the model type.
+
+      **kwargs: your inference parameters.
 
     Returns:
     --------
-      One of the clarifai.models.model_serving.models.output types. Refer to the README/docs
+      List of one of the `clarifai.models.model_serving.models.output types` or `config.inference.return_type(your_output)`. Refer to the README/docs
     """
+
     # Delete/Comment out line below and add your inference code
     raise NotImplementedError()
--- a/clarifai/models/model_serving/models/inference.py
+++ b/clarifai/models/model_serving/models/inference.py
@@ -13,7 +13,7 @@ from pathlib import Path
 from clarifai.models.model_serving.model_config import (  # noqa # pylint: disable=unused-import
     ModelTypes, get_model_config)
 
-#config = get_model_config("clarifai-model-type")
+config = get_model_config("MODEL_TYPE_PLACEHOLDER")
 
 
 class InferenceModel:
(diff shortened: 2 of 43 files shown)


clarifai/models/model_serving/models/inference.py
```diff
@@ -10,6 +10,11 @@
 import os
 from pathlib import Path
 
+from clarifai.models.model_serving.model_config import (  # noqa # pylint: disable=unused-import
+    ModelTypes, get_model_config)
+
+#config = get_model_config("clarifai-model-type")
```
ackizilkale, line 16: remove this line

ackizilkale (approved):
This is a tough PR by nature, since we have to update all these examples to handle batching. With examples comes the difficulty to maintain obviously :(

But looking good to me. Please wait for Harmit's comments also.

clarifai/models/model_serving/models/inference.py
HarmitMinhas96: It is a bit confusing that this is commented out and then replaced. Maybe leave a comment here about the behavior?

Additionally, it might be a little clearer if `"clarifai-model-type"` is replaced with something like `"MODEL_TYPE_PLACEHOLDER"`, and then the replacement code only replaces `MODEL_TYPE_PLACEHOLDER`

clarifai/models/model_serving/models/inference.py
```diff
@@ -24,19 +29,22 @@ def __init__(self) -> None:
     #self.checkpoint_path: Path = os.path.join(self.base_path, "your checkpoint filename/path")
     #self.model: Callable = <load_your_model_here from checkpoint or folder>
 
-  #Add relevant model type decorator to the method below (see docs/model_types for ref.)
-  def get_predictions(self, input_data, **kwargs):
+  #@config.inference.wrap_func
```
HarmitMinhas96, line 32: If you don't comment out the above, I don't think this needs to be commented out either. And then it wouldn't have to be replaced in `build_repository` as well

clarifai/models/model_serving/pb_model_repository.py
```diff
@@ -80,22 +82,30 @@ def build_repository(self, repository_dir: Path = os.curdir):
...
+                    f"""config = get_model_config("{self.model_config.model_type}")""")
+      x = x.replace("#@config.inference.wrap_func", "@config.inference.wrap_func")
+      return x
+
+    inference_py_path = os.path.join(model_version_path, "inference.py")
+    self._module_to_file(inference, inference_py_path, inference_insert_model_type_func)
+
     # generate test.py
+    def insert_model_type_func(x):
+      return x.replace("clarifai-model-type", self.model_config.model_type)
```
HarmitMinhas96, line 108: It would be good to also update test.py to use similarly named `MODEL_TYPE_PLACEHOLDER` and then you only need one helper function for inference.py and test.py

clarifai/models/model_serving/models/model_types.py
```diff
@@ -39,11 +39,12 @@ def parse_predictions(self, input_data: np.ndarray, *args, **kwargs):
     out_bboxes = []
     out_labels = []
     out_scores = []
-    for item in input_data:
-      preds = func(self, item, *args, **kwargs)
-      out_bboxes.append(preds.predicted_bboxes)
-      out_labels.append(preds.predicted_labels)
-      out_scores.append(preds.predicted_scores)
+    input_data = [each for each in input_data]
```
HarmitMinhas96, line 42: What's the purpose of this line? It looks like it's creating a local copy of input_data, but why? If it is needed, it can be simplified to `input_data[:]`, I think.

I also see this is done a few more times in the other wrapper definitions.

clarifai/models/model_serving/cli/deploy_cli.py
HarmitMinhas96, line 1: How does this user specify their batch size now? Should we add an arg for it? Or will that be in another PR?

HarmitMinhas96 (approved):
Looks good, mostly minor comments.

Additionally, out of scope for this ticket, but maybe we should build some CI around this to run the defined test against some generated dummy models for each model type.

clarifai/models/model_serving/pb_model_repository.py
```diff
@@ -32,24 +32,26 @@ def __init__(self, model_config: Type[TritonModelConfig]):
...
     """
     Write Python Module to file.
 
     Args:
     -----
-    module_name: Python module name to write to file
-    filename: Name of the file to write to destination_dir
-    destination_dir: Directory to save the generated triton model file.
-
+      module_name: Python module name to write to file
```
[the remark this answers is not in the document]
phatvo9: Thanks, nice catch!

clarifai/models/model_serving/models/model_types.py
phatvo9: I want to split a group of image data into individual images. This way, users can preprocess each image separately. Another reason is to ensure that the data type remains consistent with text input models.
Thank you for suggesting `input_data[:]`, looks much simpler.

clarifai/models/model_serving/models/inference.py
phatvo9: The point is `inference` imported as module, so if that line is uncommented, an error will occur since `clarifai-model-type` or `MODEL_TYPE_PLACEHOLDER` is not valid model type name. A possible solution is to read `inference` as plain text instead import it, in this case can have more flexible settings.

clarifai/models/model_serving/cli/deploy_cli.py
phatvo9: Included max_batch_size in `clarifai-model-upload-init` and updated doc

clarifai/models/model_serving/models/inference.py
HarmitMinhas96: One additional option is that `get_model_config` can just log an error/warning when `MODEL_TYPE_PLACEHOLDER` is passed in.

If you prefer the way it currently is, that's ok, but we should definitely add a comment.

clarifai/models/model_serving/models/model_types.py
HarmitMinhas96: But isn't the image data already individual images? I see it like this, but maybe I'm missing something?
```
input_data = [im1, im2, im3]
parse_predictions(input_data):
    a = input_data[:] # a is still [im1, im2, im3]
```

phatvo9: `input_data` is acquired through `get_input_tensor_by_name()` from triton request, it's numpy array with a shape of [bs, h, w, c].

clarifai/models/model_serving/models/inference.py
phatvo9: > One additional option is that get_model_config can just log an error/warning when MODEL_TYPE_PLACEHOLDER is passed in.
Thanks for suggesting, I think this approach is the optimal choice for code clarity and safety.

PR #255, opened 2023-12-26
[DEVX-322] Add Rank to PostInputsSearch

sainivedh:
## Why
 - To support Rank based search in PostInputsSearch

## Tests
 - Added `test_rank_filter_search` test case

## How
```python
from clarifai.client.search import Search
search = Search(user_id="sai_nivedh", app_id="test-imdb", top_k=3, metric="cosine")

# InputSearch by raw text
results = search.query(ranks=[{"text_raw": "Movie is Awesome"}], filters=[{"input_types": ["text"]}])

for data in results:
  for hit in data.hits:
    print(hit.input.id)
```

### DIFF
--- a/clarifai/client/search.py
+++ b/clarifai/client/search.py
@@ -18,8 +18,8 @@ from clarifai.schema.search import get_schema
 class Search(Lister, BaseClient):
 
   def __init__(self,
-               user_id,
-               app_id,
+               user_id: str,
+               app_id: str,
                top_k: int = DEFAULT_TOP_K,
                metric: str = DEFAULT_SEARCH_METRIC,
                base_url: str = "https://api.clarifai.com",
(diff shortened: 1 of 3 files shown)


clarifai/client/search.py
```diff
@@ -228,20 +236,18 @@ def query(self, ranks=[{}], filters=[{}]):
...
+                  metric=self.metric_distance)
           ])
 
       return self.list_all_pages_generator(self.STUB.PostInputsSearches,
                                            service_pb2.PostInputsSearchesRequest, request_data)
 
     # Calls PostAnnotationsSearches for annotation ranks, filters
-    rank_annot_proto, filters_annot_proto = [], []
-    for rank_dict in ranks:
-      rank_annot_proto.append(self._get_annot_proto(**rank_dict))
```
isaac-chung, line 240: How come we do not allow ranking by annotation anymore?

sainivedh: It is supported, I've shifted it to L221:225

isaac-chung: Ah my bad, I see it now.

isaac-chung (approved):

PR #256, opened 2023-12-28
[EAGLE-3792]-Test case for model upload

phatvo9:
Test dummy models for triton code


.github/workflows/run_triton_test.yaml
```diff
@@ -0,0 +1,32 @@
...
+jobs:
+  triton-model-test:
+    runs-on: ubuntu-latest
+
+    container:
+      image: nvcr.io/nvidia/tritonserver:23.10-py3
+
+    steps:
+      - name: Install dependencies
+      run: |
```
travertischio, line 17: the indentation is broken here, you either need to indent lines 17-20 to match 16 or remove 2 spaces from 16 to match 17-20 and then also fix line 22 and below

.github/workflows/run_triton_test.yaml
```diff
@@ -0,0 +1,32 @@
...
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install -r tests/requirements.txt
+
+      - name: start triton
+        run: |
+          tritonserver --model-repository ./tests/dummy_triton_models --exit-on-error false &
+        background: true
```
travertischio, line 25: this doesnt seem to actually do anything

.github/workflows/run_triton_test.yaml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install -r tests/requirements.txt
+
+      - name: start triton
+        run: |
+          tritonserver --model-repository ./tests/dummy_triton_models --exit-on-error false &
```
travertischio, line 24: this can just be `run: < command >`
theres no need for the multi line syntax

.github/workflows/run_triton_test.yaml
```diff
@@ -0,0 +1,32 @@
...
+        background: true
+
+      - name: Wait for Triton to Initialize
+        run: sleep 10
+
+      - name: Start Model Tests
+        run: |
+          python3 -m pytest tests/dummy_triton_models/_test_all_dummies.py -s
```
travertischio, line 32: this can just be `run: < command >`
theres no need for the multi line syntax

ackizilkale:
This is necessarily a huge PR. Went through it, did not find anything wrong. 

@phatvo9 how long do you think these tests will take to run?

.github/workflows/run_triton_test.yaml
phatvo9: Thanks for spotting that error

phatvo9:
> This is necessarily a huge PR. Went through it, did not find anything wrong.
> 
> @phatvo9 how long do you think these tests will take to run?

With these dummy models, I think the test won't take more than 15 seconds

.github/workflows/run_triton_test.yaml
```diff
@@ -0,0 +1,29 @@
...
+  pull_request:
+    paths:
+      - 'clarifai/models/**'
+
+jobs:
+  triton-model-test:
+
+    runs-on: ubuntu-latest
+    container:
+      image: nvcr.io/nvidia/tritonserver:23.10-py3
```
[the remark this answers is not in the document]
phatvo9: I want to make it's as simple as possible so anyone can do test

ackizilkale (approved):
lgtm

PR #262, opened 2024-01-05
[LT-4894] Add RAG class

isaac-chung:
**Beta**
New RAG class to streamline chat experience with a single interface.
Left the other methods as TODO so that other can work on it in parallel.

Tested manually. Note that rag-prompter is only public in dev at the moment. Can test with "prompter" model type.

### DIFF
--- a/clarifai/rag.py
+++ b/clarifai/rag.py
@@ -1,0 +1,122 @@
+from datetime import datetime
+
+import yaml
+from google.protobuf.struct_pb2 import Struct
+
+from clarifai.client.app import App
+from clarifai.client.model import Model
+from clarifai.client.user import User
+from clarifai.client.workflow import Workflow
+
+
+class RAG:
+  """
+    RAG is a class for Retrieval Augmented Generation.
+
+    Example:
+        >>> from clarifai.rag import RAG
+        >>> rag_agent = RAG()
+    """
+
+  def __init__(self,
+               workflow_url: str = None,
+               base_url: str = "https://api.clarifai.com",
+               pat: str = None,
+               **kwargs):
+    """
+        Initialize an empty or existing RAG.
+        """
+    w = Workflow(url=workflow_url, base_url=base_url, pat=pat)
+    self._prompt_workflow = w
+    self._app = App(app_id=w.app_id, base_url=w.base, pat=w.pat)
+    del w
+
+  def setup(self,
+            user_id: str = None,
+            llm_url: str = "https://clarifai.com/mistralai/completion/models/mistral-7B-Instruct",
+            base_workflow: str = "Text",
+            workflow_yaml_filename: str = 'prompter_wf.yaml',
+            base_url: str = "https://api.clarifai.com",
+            pat: str = None,
+            **kwargs):
+    """
+        Creates an app with `Text` as base workflow, create prompt model, create prompt workflow.
+        Example:
+        >>> from clarifai.rag import RAG
+        >>> rag_agent = RAG().setup()
+        """
+    user = User(user_id=user_id, base_url=base_url, pat=pat)
+    llm = Model(llm_url)
+
+    params = Struct()
+    params.update({
+        "prompt_template":
+            "Context information is below:\n{data.hits}\nGiven the context information and not prior knowledge, answer the query.\nQuery: {data.text.raw}\nAnswer: "
+    })
+    prompter_model_params = {"params": params}
+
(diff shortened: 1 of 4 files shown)


clarifai/rag.py
```diff
@@ -0,0 +1,122 @@
...
+               **kwargs):
+    """
+        Initialize an empty or existing RAG.
+        """
+    w = Workflow(url=workflow_url, base_url=base_url, pat=pat)
+    self._prompt_workflow = w
+    self._app = App(app_id=w.app_id, base_url=w.base, pat=w.pat)
+    del w
+
+  def setup(self,
```
sainivedh, line 34: We could have this as a `classmethod` instead
```python
rag = RAG.setup()
```

clarifai/rag.py
```diff
@@ -0,0 +1,122 @@
...
+            user_id: str = None,
+            llm_url: str = "https://clarifai.com/mistralai/completion/models/mistral-7B-Instruct",
+            base_workflow: str = "Text",
+            workflow_yaml_filename: str = 'prompter_wf.yaml',
```
sainivedh, line 38: any reason on why we are restricting users to provide their yml(`prompt_template`) ?.

clarifai/rag.py
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+    with open(workflow_yaml_filename, 'w') as out_file:
+      yaml.dump(workflow_dict, out_file, default_flow_style=False)
+
+    ## Create prompt workflow
+    self._prompt_workflow = app.create_workflow(config_filepath=workflow_yaml_filename)
+    self._app = app
+
+    del user, llm, prompter_model, prompter_model_params
```
sainivedh, line 99: Maybe add a helper function in [urls](https://github.com/Clarifai/clarifai-python/blob/master/clarifai/urls/helper.py), that generates urls for resource instance.

workflow_url = ClarifaiURLHelper.get_clarifai_url(self._prompt_workflow)

```suggestion
    return cls(workflow_url, base_url, pat)
```

clarifai/rag.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class RAG:
+  """
+    RAG is a class for Retrieval Augmented Generation.
+
+    Example:
+        >>> from clarifai.rag import RAG
+        >>> rag_agent = RAG()
+    """
```
sainivedh, line 19: Doc strings indentation is inconsistent with the rest of SDK

sainivedh (commented):
Could we organize this by placing it in a dedicated "rag" folder?
As we expand, there could be a need for utilities specific to rag functionalities.

```
clarifai
   | client
   | datasets
   | rag
       | __init__.py 
       | rag.py
       | utils.py
```

clarifai/rag.py
isaac-chung: Pre-commit hooks all passed.

clarifai/rag.py
isaac-chung: Not sure if this comment could be made more clear to help answer this question:
> \## Generate a tmp yaml file for workflow creation

Maybe we could specify a `/tmp` location or something. I was concerned about permissions.

clarifai/rag.py
isaac-chung: This assumes `setup` to be a classmethod. Let's merge the conversation to just the one above.

clarifai/rag.py
sainivedh: Yes, but instead of `workflow_yaml_filename` maybe we could expose `prompt_template` arg ?

```
"prompt_template":
            "Context information is below:\n{data.hits}\nGiven the context information and not prior knowledge, answer the query.\nQuery: {data.text.raw}\nAnswer: "
    }
```

isaac-chung: Those are separate things. Notice that the workflow yaml file is _generated_, not user supplied.
I think at this stage we have no need to expose the `prompt_template`.

clarifai/rag.py
sainivedh: > init would take care of setting up a new RAG setup

 `setup()` takes `base_workflow`, `llm_url` etc as args. Not sure if we could merge with init and make `from_workflow(url)` work

isaac-chung: I think I get what you mean now. Will try out your suggestion

sainivedh (approved):
Looks Correct

PR #263, opened 2024-01-08
[DEVX-324]: Download inputs

sanjaychelliah:
# what

- Download inputs feature(SDH)
```python
#listing inputs
input_generator = input_obj.list_inputs(page_no=1,per_page=1,input_type='image')
inputs_list = list(input_generator)

#downloading_inputs
input_bytes = input_obj.download_inputs(inputs_list)
with open('demo.jpg','wb') as f:
  f.write(input_bytes[0])
```
- Unit test and readme changes for the same.

### DIFF
--- a/clarifai/client/input.py
+++ b/clarifai/client/input.py
@@ -7,11 +7,13 @@ from concurrent.futures import ThreadPoolExecutor, as_completed
 from multiprocessing import cpu_count
 from typing import Generator, List, Union
 
+import requests
 from clarifai_grpc.grpc.api import resources_pb2, service_pb2  # noqa: F401
 from clarifai_grpc.grpc.api.resources_pb2 import Annotation, Audio, Image, Input, Text, Video
 from clarifai_grpc.grpc.api.status import status_code_pb2, status_pb2
 from google.protobuf.json_format import MessageToDict
 from google.protobuf.struct_pb2 import Struct
+from requests.adapters import HTTPAdapter, Retry
 from tqdm import tqdm
 
 from clarifai.client.base import BaseClient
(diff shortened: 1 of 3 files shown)


clarifai/client/input.py
```diff
@@ -729,6 +731,46 @@ def delete_inputs(self, inputs: List[Input]) -> None:
...
+        if response.status_code == 200:
+          final_inputs.append(response.content)
+      if input.data.audio.url:
+        response = session.get(input.data.audio.url, stream=True)
+        if response.status_code == 200:
+          final_inputs.append(response.content)
+      if input.data.text.url:
+        response = session.get(input.data.text.url, stream=True)
+        if response.status_code == 200:
+          final_inputs.append(response.content)
```
sainivedh, line 770: I think we can refactor this down to

```python
# download inputs
data_types = ['image', 'video', 'audio', 'text']
for input in inputs:
    for data_type in data_types:
        url = getattr(input.data, data_type).url
        if url:
            response = session.get(url, stream=True)
            if response.status_code == 200:
                final_inputs.append(response.content)
```

clarifai/client/input.py
```diff
@@ -729,6 +731,46 @@ def delete_inputs(self, inputs: List[Input]) -> None:
       raise Exception(response.status)
     self.logger.info("\nInputs Deleted\n%s", response.status)
 
+  def download_inputs(self, inputs: List[Input]) -> List[bytes]:
+    """Download list of input objects from the app.
```
sainivedh, line 735: Shouldn't we store inputs as files, instead of returning `bytes` ?

sanjaychelliah: This is for the user to use the file as per their need, they can use it in the runtime or choose to save the file to the local. Like how we have in model prediction..

clarifai/client/input.py
```diff
@@ -729,6 +731,37 @@ def delete_inputs(self, inputs: List[Input]) -> None:
...
+    retries = Retry(total=3, backoff_factor=0.1, status_forcelist=[500, 502, 503, 504])
+    session.mount('https://', HTTPAdapter(max_retries=retries))
+    session.headers.update({'Authorization': self.metadata[0][1]})
+    # download inputs
+    data_types = ['image', 'video', 'audio', 'text']
+    for input in inputs:
+      for data_type in data_types:
+        url = getattr(input.data, data_type).url
+        if url:
+          response = session.get(url, stream=True)
```
sainivedh, line 759: As this is io intensive, we can make use of `ThreadPoolExecutor`

sanjaychelliah: For bulk download, we have dataset export function, there we use Threadpoolexecutor. This is majorly for minimal input download for quick viewing/storing. I think its not needed here. WDYT?

sainivedh (approved):

PR #264, opened 2024-01-08
Add RAG chat method

isaac-chung:
## What
The chat method for RAG follows the openAI API format, i.e. it [expects messages as a list of dicts](https://docs.anthropic.com/claude/reference/messages_post) like:
```
[
          {"role": "user", "content": "Hello there."},
          {"role": "assistant", "content": "Hi, I'm Claude. How can I help you?"},
          {"role": "user", "content": "Can you explain LLMs in plain English?"},
]
```
This aligns with most if not all cloud LLM providers in terms of format.

### Notes

- Users can choose whether to manage chat state themselves or let us do it.
- Added workflow state support in workflow predict
- work is in progress for the actual backend change for `client_manage_state=false`. May consider adding "not implemented" for the moment and a test XFAIL

### DIFF
--- a/clarifai/client/workflow.py
+++ b/clarifai/client/workflow.py
@@ -75,7 +75,7 @@ class Workflow(Lister, BaseClient):
         output_config=self.output_config)
 
     if workflow_state_id:
-      request.workflow_state = resources_pb2.WorkFlowState(id=workflow_state_id)
+      request.workflow_state.id = workflow_state_id
 
     start_time = time.time()
     backoff_iterator = BackoffIterator()
--- a/clarifai/client/workflow.py
+++ b/clarifai/client/workflow.py
@@ -63,6 +63,7 @@ class Workflow(Lister, BaseClient):
 
     Args:
         inputs (list[Input]): The inputs to predict.
+        workflow_state_id (str): key for the workflow state cache that stores the workflow node predictions.
     """
     if len(inputs) > MAX_WORKFLOW_PREDICT_INPUTS:
       raise UserError(f"Too many inputs. Max is {MAX_WORKFLOW_PREDICT_INPUTS}."
--- a/clarifai/client/workflow.py
+++ b/clarifai/client/workflow.py
@@ -58,7 +58,7 @@ class Workflow(Lister, BaseClient):
     BaseClient.__init__(self, user_id=self.user_id, app_id=self.app_id, base=base_url, pat=pat)
     Lister.__init__(self)
 
-  def predict(self, inputs: List[Input]):
+  def predict(self, inputs: List[Input], workflow_state_id: str = None):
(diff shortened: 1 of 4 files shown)


clarifai/client/workflow.py
```diff
@@ -58,7 +58,7 @@ def __init__(self,
     BaseClient.__init__(self, user_id=self.user_id, app_id=self.app_id, base=base_url, pat=pat)
     Lister.__init__(self)
 
-  def predict(self, inputs: List[Input]):
+  def predict(self, inputs: List[Input], workflow_state_id: str = None):
     """Predicts the workflow based on the given inputs.
 
     Args:
```
sanjaychelliah, line 64: Can we add a description about **_workflow_state_id_** in args.

clarifai/rag/rag.py
```diff
@@ -120,10 +126,41 @@ def upload():
...
     """
-    pass
+    if client_manage_state:
+      single_prompt = convert_messages_to_str(messages)
+      input_proto = Inputs._get_proto("", "", text_pb=resources_pb2.Text(raw=single_prompt))
+      response = self._prompt_workflow.predict([input_proto])
+      messages.append(format_assistant_message(response.results[0].outputs[-1].data.text.raw))
+      return messages
+
+    # server-side state management
```
sanjaychelliah, line 152: Is it possible to get `message` as a string when its server-side state management. So that It'll be easy for the user to just give the chat as string and not as List(Dict) for ease of use.

For examples,
When its client manage state,
`rag_agent.chat([{"role": "user", "content": "Hello there."}], client_manage_state = True)`

and when server-side state management, just the string like,
`rag_agent.chat("Hello there.")`

sanjaychelliah (commented):
LGTM, Few minor comments

clarifai/client/workflow.py
isaac-chung: For sure.

clarifai/rag/rag.py
isaac-chung: That's a good point. And for server-side state management, would the output be a `str` as well instead of `List[dict]`?

I am curious about what @zhangxiangxiao thinks. Our initial discussions pointed us to follow OpenAI API format such that this can be used as a drop-in replacement.

sanjaychelliah: Yeah,  I think that may be a good experience, output also as `str` in server-side state management.

isaac-chung: Another application for output as `str` would be for some sort of cli chat, like how [ollama](https://ollama.ai/) has it. But since we are providing an API, we wish to follow the OpenAI format for the above reason (as drop-in replacement). I think we can consider output as `str` for future extensions.

sanjaychelliah (approved):

PR #270, opened 2024-01-16
DEVX-330 - modified process_response_keys function to support metadata fields

mogith-pn:
**Description :**

list_inputs function is not fetching metadata field since the process_response_keys function which used to convert the input response into object omits it. 

**Changes made :** 

- Modified the function to include metadata fields as well. 
- Tested with other functions which call this internally such as, list_workflows, list_apps, list_datasets, list_modules, list_models, list_annotations, list_concepts.

https://clarifai.atlassian.net/browse/DEVX-330

### DIFF
--- a/clarifai/client/base.py
+++ b/clarifai/client/base.py
@@ -119,7 +119,12 @@ class BaseClient:
             value_s.update(value)
             value = value_s
           elif key in ['metadata']:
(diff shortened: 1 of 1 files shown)


sanjaychelliah (approved):
LGTM

PR #275, opened 2024-01-19
Support case when existing app without workflow is provided

isaac-chung:
## Problem
When users have an existing app with inputs, they currently do not have a way to start RAG via the RAG class. The setup method supports creating everything, including the app, from scratch. But it is possible that the data ingestion phase is too large to be replicated.

## Improvement
Add a parameter and checks in the `setup` method. Preview:
```python
from clarifai.rag import RAG
rag_agent = RAG.setup(app_url=YOUR_APP_URL)
rag_agent.chat(messages=[{"role":"human", "content":"What is Clarifai"}])
```

### DIFF
--- a/clarifai/rag/rag.py
+++ b/clarifai/rag/rag.py
@@ -25,6 +25,7 @@ class RAG:
     Example:
         >>> from clarifai.rag import RAG
         >>> rag_agent = RAG(workflow_url=YOUR_WORKFLOW_URL)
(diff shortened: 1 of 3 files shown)


clarifai/rag/rag.py
sanjaychelliah, line 1: Can we add check to catch the scenario in which both `user_id` and `app_url` are given by mistake.

isaac-chung: Yep

sanjaychelliah (approved):
LGTM 👍

PR #280, opened 2024-01-29
use kwargs in RAG setup

isaac-chung:
Allow specifying parameters like `min_score` via kwargs in RAG setup method.


sanjaychelliah (approved):
LGTM 👍

PR #281, opened 2024-01-29
Updated xview dataloader function with task function

mogith-pn:
- Updated the xview dataloader function with task function.


mogith-pn:
> LGTM. Can you make the lint change thats failing..

sure.

PR #287, opened 2024-02-01
[EAGLE-3932] model uploadv2 doc

phatvo9:
### What
Update doc of model-serving


ackizilkale (approved):
looks great. Looking forward to using this now.

PR #290, opened 2024-02-07
[EAGLE-3919] - eval endpoints 

phatvo9:
### What
Bring gRPC eval endpoints to SDK

Note: We currently don't have endpoint to retrieve eval metrics of specific model type

### DIFF
--- a/clarifai/client/model.py
+++ b/clarifai/client/model.py
@@ -1,6 +1,6 @@
 import os
 import time
-from typing import Any, Dict, Generator, List
+from typing import Any, Dict, Generator, List, Union
 
 import requests
 import yaml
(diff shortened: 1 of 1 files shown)


clarifai/client/model.py
```diff
@@ -576,3 +576,179 @@ def __str__(self):
...
+    Raises:
+        Exception: Failed to call API
+
+    Returns:
+        List[resources_pb2.EvalMetrics]: list of eval_metrics
+    """
+    request = service_pb2.ListModelVersionEvaluationsRequest(
+        user_app_id=self.user_app_id,
+        model_id=self.id,
+        model_version_id=self.model_info.model_version.id)
```
sanjaychelliah, line 592: Can we have a check and run model.load_info() if version id is none.. Usually model version ID will not be there..

clarifai/client/model.py
```diff
@@ -576,3 +576,179 @@ def __str__(self):
...
+    )
+    request = service_pb2.PostEvaluationsRequest(
+        user_app_id=self.user_app_id,
+        eval_metrics=[eval_metric],
+    )
+    response = self._grpc_request(self.STUB.PostEvaluations, request)
+    if response.status.code != status_code_pb2.SUCCESS:
+      raise Exception(response.status)
+
+    return response.eval_metrics
```
sanjaychelliah, line 672: Is it possible to parse the response proto and make it as a dict with only the metric details.It will be a good user experience. WDYT?

sanjaychelliah (commented):
Great Addition @phatvo9. 🙌  Adding some comments..

clarifai/client/model.py
phatvo9: In this endpoint wrapper, I aim for users to be able to utilize type hints and parse the output data easily. The output of ListEval includes the evaluation ID, dataset information, summary, and status of evaluations. This information is crucial for further work, such as providing the evaluation ID to retrieve detailed evaluation results or checking if a dataset has been evaluated.
I plan to create some helper functions based on these wrappers to make users' lives easier.

clarifai/client/model.py
phatvo9: You're right, I should check it. I believe I can use an assert statement to check if model_version.id is None?

clarifai/client/model.py
sanjaychelliah: Got it 👍

sanjaychelliah (approved):

PR #299, opened 2024-02-21
[DEVX-346]: Requirements Update

sanjaychelliah:
### Why

- To make the clarifai install less bloated

### How

- Removed pandas, opencv and replaced with pillow
- Removed coco_segmentation loader since it was using cv2 functionalities( COCO Segmentation loader was already added in **clarifai-datautils**)
- Moved llama-index,pypdf for RAG utilites( load documents, split documents) as an optional import for users who are uploading data using RAG

### Notes

- RAG utilites is a temporary fix, data utilities will be permanently moved to **clarifai-datautils** soon, after that,this dependency on external libraries for RAG utils will be taken out.


sainivedh (approved):
lgtm

PR #303, opened 2024-03-05
updates for data export

deigen:
Standalone function to get the request and get an archive zip url with input protos, that we can use to iterate inputs with annos and convert to other formats.

### DIFF
--- a/clarifai/client/dataset.py
+++ b/clarifai/client/dataset.py
@@ -169,6 +169,12 @@ class Dataset(Lister, BaseClient):
       }
       yield Dataset.from_auth_helper(self.auth_helper, **kwargs)
 
+  def iter_inputs(self):
+    return iter(DatasetExportReader(archive_url=self.archive_zip()))
+
+  def __iter__(self):
+    return self.iter_inputs()
+
   def _concurrent_annot_upload(self, annots: List[List[resources_pb2.Annotation]]
                               ) -> Union[List[resources_pb2.Annotation], List[None]]:
     """Uploads annotations concurrently.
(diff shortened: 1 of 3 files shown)


sanjaychelliah (approved):

clarifai/datasets/export/inputs_annotations.py
```diff
@@ -59,18 +61,21 @@ def __init__(self,
   def _download_temp_archive(self, archive_url: str,
                              chunk_size: int = 128) -> tempfile.TemporaryFile:
     """Downloads the temp archive of InputBatches."""
-    r = self.session.get(archive_url, stream=True)
+    session = requests.Session()
+    r = session.get(archive_url, stream=True)
```
deigen, line 65: This may need to use self.session, which is constructed with the auth headers outside this class.  I'd like a better way to set the auth headers than passing in a session, maybe using a global (or per-thread) context, so that we don't have to construct a session in the caller.  But right now I think self.session may be needed here.

sanjaychelliah: Agreed. But here the session with auth headers is giving an error for the zip archive whereas its working without auth headers. Thats why not used the session with auth headers for now..

clarifai/client/dataset.py
```diff
@@ -169,6 +169,12 @@ def list_versions(self, page_no: int = None,
       }
       yield Dataset.from_auth_helper(self.auth_helper, **kwargs)
 
+  def iter_inputs(self):
+    return iter(DatasetExportReader(archive_url=self.archive_zip()))
+
+  def __iter__(self):
+    return self.iter_inputs()
+
```
isaac-chung, line 177: This could be merged into the `__iter__` method:
```python
def __iter__(self):
  return iter(DatasetExportReader(archive_url=self.archive_zip()))
```

sanjaychelliah: Modified in this [PR](https://github.com/Clarifai/clarifai-python/pull/315) 👍

PR #317, opened 2024-03-20
DEVX-335 limited batch size for inputs upload 

mogith-pn:
### Why

* To limit max upload batch size for `upload_inputs` function. 
* Although the integrations and others directly chunking and uploading the inputs, this is to restrict users from directly using `upload_inputs` function with higher batch size > 128.

* Also this PR intends to fix the below model version id parameter inconsistency error, 

image
image
Reference ticket - https://clarifai.atlassian.net/browse/DEVX-369
### How

* Standardised the model version id parameter in both `app.model()` class as well in `model.__init__()`.

### Tests

* Normal tests are passing and tested locally with all the supplement functions that can be used with app.model()


clarifai/client/input.py
```diff
@@ -19,6 +19,7 @@
 from clarifai.client.base import BaseClient
 from clarifai.client.lister import Lister
 from clarifai.constants.dataset import MAX_RETRIES
+from clarifai.constants.rag import MAX_UPLOAD_BATCH_SIZE
```
sanjaychelliah, line 22: we can add this constant in dataset itself to avoid confusion..
`from clarifai.constants.dataset import MAX_RETRIES, MAX_UPLOAD_BATCH_SIZE `

mogith-pn: > we can add this constant in dataset itself to avoid confusion.. `from clarifai.constants.dataset import MAX_RETRIES, MAX_UPLOAD_BATCH_SIZE `

Even I thought of same, should we do that or we can add new constant file for input.py ?? 
WDYT

sanjaychelliah: Yeah constants for input.py is more relevant 👍

sanjaychelliah (approved):
LGTM!

PR #319, opened 2024-03-26
[DEVX-400] Add root cert support

sainivedh:
### Why

* Adding root certificate support to establish secure gRPC connections.

### How

* Adding root certificate param in all the classes and auth helper
* Updating the grpc to latest version

### Tests

* Added a test to validate the root certificate path in test_auth.


sanjaychelliah (approved):
LGTM

PR #320, opened 2024-03-26
Add VERSION and requirements.txt to setup.py

markmc:
These files aren't currently being included in the sdist, giving e.g.

```
$ python setup.py --help
...
FileNotFoundError: [Errno 2] No such file or directory: 'VERSION'
```

Looks like this has been an issue since 9.9.1.

This also means you can't use the more recent `pypa/build` module to build a wheel for this package:

```
$ python -m build
* Creating isolated environment: venv+pip...
* Installing packages in isolated environment:
  - setuptools >= 40.8.0
* Getting build dependencies for sdist...
* Building sdist...
...
copying files to clarifai-10.2.1...
copying LICENSE -> clarifai-10.2.1
copying MANIFEST.in -> clarifai-10.2.1
copying README.md -> clarifai-10.2.1
copying pyproject.toml -> clarifai-10.2.1
copying setup.py -> clarifai-10.2.1
...
* Building wheel from sdist
* Creating isolated environment: venv+pip...
* Installing packages in isolated environment:
  - setuptools >= 40.8.0
* Getting build dependencies for wheel...
Traceback (most recent call last):
...
FileNotFoundError: [Errno 2] No such file or directory: 'VERSION'

ERROR Backend subprocess exited when trying to invoke get_requires_for_build_wheel
```

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
 global-include *.css
 include clarifai/modules/style.css
 recursive-include clarifai *
+include VERSION requirements.txt


markmc:
Consider replacing [this](https://github.com/Clarifai/clarifai-python/blob/13b452ca189bc79f9f1918569f75ec92617bb7a8/.github/workflows/publish.yml#L29) with `python -m build` ?

sanjaychelliah (approved):

sanjaychelliah:
> Consider replacing [this](https://github.com/Clarifai/clarifai-python/blob/13b452ca189bc79f9f1918569f75ec92617bb7a8/.github/workflows/publish.yml#L29) with `python -m build` ?

Thanks for the contribution. 

Adding those missing files in MANIFEST.in will solve this. So is there a specific reason to replace [this](https://github.com/Clarifai/clarifai-python/blob/13b452ca189bc79f9f1918569f75ec92617bb7a8/.github/workflows/publish.yml#L29) with `python -m build`?

markmc:
> > Consider replacing [this](https://github.com/Clarifai/clarifai-python/blob/13b452ca189bc79f9f1918569f75ec92617bb7a8/.github/workflows/publish.yml#L29) with `python -m build` ?
> 
> Thanks for the contribution.
> 
> Adding those missing files in MANIFEST.in will solve this. So is there a specific reason to replace [this](https://github.com/Clarifai/clarifai-python/blob/13b452ca189bc79f9f1918569f75ec92617bb7a8/.github/workflows/publish.yml#L29) with `python -m build`?

That's a good question, sorry I didn't clarify that!

The difference is that an sdist would first be built, and then a wheel built from the sdist. As [the docs](https://build.pypa.io/en/stable/) say:

```
By default, a source distribution (sdist) is built from {srcdir} and a binary distribution (wheel) is built from the sdist. This is recommended as it will ensure the sdist can be used to build wheels
```

So using this approach, the issue would have been found immediately because 'python -m build' would have failed

sanjaychelliah:
> > > Consider replacing [this](https://github.com/Clarifai/clarifai-python/blob/13b452ca189bc79f9f1918569f75ec92617bb7a8/.github/workflows/publish.yml#L29) with `python -m build` ?
> > 
> > 
> > Thanks for the contribution.
> > Adding those missing files in MANIFEST.in will solve this. So is there a specific reason to replace [this](https://github.com/Clarifai/clarifai-python/blob/13b452ca189bc79f9f1918569f75ec92617bb7a8/.github/workflows/publish.yml#L29) with `python -m build`?
> 
> That's a good question, sorry I didn't clarify that!
> 
> The difference is that an sdist would first be built, and then a wheel built from the sdist. As [the docs](https://build.pypa.io/en/stable/) say:
> 
> ```
> By default, a source distribution (sdist) is built from {srcdir} and a binary distribution (wheel) is built from the sdist. This is recommended as it will ensure the sdist can be used to build wheels
> ```
> 
> So using this approach, the issue would have been found immediately because 'python -m build' would have failed

Thank you. 👍

PR #326, opened 2024-04-05
[EAGLE-4139] add fix to stop infra alerts

wemoveon2:
### Why

* One of the first steps of training is a query that throws an error if the app doesn't exist. We're deleting the app so quickly that this step happens after the app is deleted, so the training pod does a non 0 exit, which triggers infra alerts, which triggers SREs to trigger me.

### How

* Sleep for a lil before deleting everything so we get past the first step. 

### Tests

*

### Notes

*

### DIFF
--- a/tests/test_model_train.py
+++ b/tests/test_model_train.py
@@ -1,6 +1,7 @@
 import logging
 import os
 from datetime import datetime
+import time
 
 import pytest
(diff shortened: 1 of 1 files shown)


sanjaychelliah (approved):
👍

PR #329, opened 2024-04-08
update upload cli doc

phatvo9:
### What

Update upload model cli doc


clarifai/models/model_serving/README.md
```diff
@@ -142,6 +142,8 @@ Upload
 
 ```bash
 $ clarifai upload model --url <url> --user-app <your_user_id>/<your_app_id> --id <your_model_id>
+# or upload built file directly
```
ackizilkale, line 145: I think we should put this on top and url as the second option

ackizilkale: Cause I believe most users will use direct upload. A very small minority will be interested in url

ackizilkale (approved):

PR #332, opened 2024-04-18
[DEVX-412]: RAG model_id bug

sanjaychelliah:
### Why

* RAG.setup() fails if we delete a specific workflow and give the same id.
* [RAG.upload() to support folder of text files.
](https://clarifai.atlassian.net/browse/DEVX-393?atlOrigin=eyJpIjoiMTMxZTE5ZDMwMDMyNDZhOWI5ODRlN2QwZmJhNjU1MjMiLCJwIjoiaiJ9)

### DIFF
--- a/clarifai/rag/rag.py
+++ b/clarifai/rag/rag.py
@@ -110,7 +110,7 @@ class RAG:
     prompter_model_params = {"params": params}
 
     ## Create rag-prompter model and version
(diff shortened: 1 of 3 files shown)


isaac-chung:
While adding a timestamp may solve the issue temporarily, I'd suggest finding the root cause. Having multiple workflows with suffixes like 74927384 is not great UX. Happy to continue the discussion on slack.

isaac-chung (approved):

PR #339, opened 2024-04-29
DEVX-421-modified to catch detailed failed annotations status

mogith-pn:
### Why

* PR opened with reference to https://clarifai.atlassian.net/browse/DEVX-421
*  Failed annotations are not capturing the full details of failure in the log.

### Tests

* Not needed


sanjaychelliah (approved):
👍

PR #342, opened 2024-05-07
DEVX-426: get_upload_status overriding log_warnings table in log file issue

mogith-pn:
### Why

* This PR is in reference to the ticket --> https://clarifai.atlassian.net/browse/DEVX-426.

### How

* Issue: When trying `get_upload_status` args along with `log_warnings` in `dataset.upload_dataset` function we are facing an issue where the upload status table overrides the log file since `self.logger` is shared between both loggers.

### Tests

* No tests, Tested locally with all parameters and batch size to ensure it fixes the problem.


mogith-pn:
@sanjaychelliah @sainivedh 
could you please look into this ?

sanjaychelliah (approved):
👍

PR #346, opened 2024-06-13
[EAGLE-4453] Stream support

luv-bansal:
### What 

- Support streaming endpoints GenerateModelOutputs or StreamModelOutputs from the stub
-  Easier to setup client from env and just use it's stub like: 
```
from clarifai.client import BaseClient
client = BaseClient.from_env()

# advanced grpc users can use the stub directly from there. 
stub = client.Stub
response = stub.PostModelOutputs(...)
```
- Remove validation of app_id   as the base client

### Why

*

### How

*

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/client/auth/stub.py
+++ b/clarifai/client/auth/stub.py
@@ -1,3 +1,4 @@
+import itertools
 import logging
 import time
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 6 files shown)


zeiler (approved):
Looks good!

clarifai/client/auth/stub.py
```diff
@@ -20,6 +21,34 @@
...
+
+  # Check if the response is an instance of a gRPC streaming call
+  if isinstance(response, grpc._channel._MultiThreadedRendezvous):
+    try:
+      # Check just the first response in the stream for validation
+      first_res = next(response)
+      validated_response = handle_simple_response(first_res)
+      if validated_response is not None:
+        # Have to return that first response and the rest of the stream.
+        return itertools.chain([validated_response], response)
```
zeiler, line 42: @luv-bansal see here where I had to change the logic so that it includes the first chunk off the stream. Tested that this works and will merge now.

zeiler (approved):

clarifai/client/auth/stub.py
luv-bansal: Yeah, this seems good.

PR #351, opened 2024-06-20
Update text.py

sanjaychelliah:
### Why

* Uploading data to a dataset which has already inputs in it fails.

### How

* Adding random ID as in Image features


mogith-pn (approved):

mogith-pn:
LGTM !

PR #352, opened 2024-06-20
VERSION 10.5.2

sanjaychelliah:
### Changed
- Text Features to add random ID as input if input ID is not provided in Dataloader  [(#351)](https://github.com/Clarifai/clarifai-python/pull/351)


mogith-pn (approved):
LGTM !

PR #353, opened 2024-06-26
[EAGLE-4531]: 10.5.3 and avoid writing to dirs outside of CLI

zeiler:
Can't write to disk in imported functions as some file systems might be protected from writes. This is only needed for CLI usage so move to that part of the package.


clarifai/models/model_serving/cli/login.py
```diff
@@ -1,6 +1,6 @@
 import argparse
 
-from clarifai.utils.constants import CLARIFAI_PAT_PATH
+from clarifai.model_serving.constants import CLARIFAI_PAT_PATH
```
wemoveon2, line 3: should this be `clarifai.models.model_serving.constants`?

clarifai/versions.py
```diff
@@ -1,6 +1,6 @@
 import os
 
-CLIENT_VERSION = "10.5.2"
+CLIENT_VERSION = "10.5.3"
```
wemoveon2, line 3: maybe this should read `VERSION` instead of hardcoding?

clarifai/models/model_serving/cli/login.py
wemoveon2: same with a few of the other imports

wemoveon2 (approved):
adjacent comments, can be ignored since it seems tests pass? or do we not have coverage?

clarifai/models/model_serving/cli/login.py
zeiler: good catch!

clarifai/versions.py
zeiler: yeah I find this weird too. maybe @sanjaychelliah can take up

VERSION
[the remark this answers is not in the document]
zeiler: yeah agreed, we do this in the grpc ones already.

clarifai/models/model_serving/cli/login.py
wemoveon2: oof... did we not have test coverage of this...???

PR #358, opened 2024-07-08
[DEVX-394]: Add Env Vars to RAG

srikanthbachala20:
### Why

* https://clarifai.atlassian.net/browse/DEVX-394

### How

* Get user_id form env vars using utils function


sanjaychelliah (approved):
👍

PR #366, opened 2024-07-31
DEVX-326 - Removed PostAnnotationsSearches from SDK search to avoid duplicated results

mogith-pn:
### Why

* https://clarifai.atlassian.net/browse/DEVX-326 - Using 2 types of search request from SDK potentially blocks the users to use different types of filters altogether. Filters like `metadata`, `concepts` , `geo-location` can't be used along with `input_types`, `input_dataset_ids` etc due to former comes under `Annotation` proto and latter comes under `Input` proto.

* https://clarifai.atlassian.net/browse/DEVX-546 - Using `PostAnnotationsSearches` also returns duplicated triplets due to some default reasons, which is listed [here](https://clarifai.slack.com/archives/CK3QXM78Q/p1701841567690739) . 

### How

* By appropriately packing the filters with their associate proto and passing it to `PostInputsSearches` function to achieve this.

Screenshot 2024-07-31 at 11 29 58 PM

### Tests

* Added a test function inside tests/test_search.py named `test_rank_search_deduplicate` to check if the request is returning duplicates in the response.

* Further tested the function extensively in local notebooks with different filter combinations.

### DIFF
--- a/clarifai/client/search.py
+++ b/clarifai/client/search.py
@@ -277,7 +277,8 @@ class Search(Lister, BaseClient):
         Get successful inputs of type image or text
         >>> from clarifai.client.search import Search
         >>> search = Search(user_id='user_id', app_id='app_id', top_k=10, metric='cosine')
-        >>> res = search.query(filters=[{'input_types': ['image', 'text']}, {'input_status_code': 30000}])
+        # This performs OR operation on input_types and input_status_code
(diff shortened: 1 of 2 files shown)


nitinbhojwani (approved):

sanjaychelliah (approved):
👍

PR #371, opened 2024-08-12
[DEVX-538]: Addition of Concept Relation Operations

srikanthbachala20:
### Why

* To support Create, List and Delete Concept Relations for a App
* Ref : https://clarifai.atlassian.net/browse/DEVX-538

### How

* Created a function each to create, list and delete concept relations for the app
* Also added a function to add concepts for the app

### Tests

* Added tests for creation, listing and deletion of concept relations of a app in tests/test_app.py

### DIFF
--- a/clarifai/client/app.py
+++ b/clarifai/client/app.py
@@ -4,7 +4,7 @@ from typing import Any, Dict, Generator, List
 
 import yaml
 from clarifai_grpc.grpc.api import resources_pb2, service_pb2
-from clarifai_grpc.grpc.api.resources_pb2 import Concept
+from clarifai_grpc.grpc.api.resources_pb2 import Concept, ConceptRelation
 from clarifai_grpc.grpc.api.status import status_code_pb2
 from google.protobuf.json_format import MessageToDict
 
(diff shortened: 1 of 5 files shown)


clarifai/client/app.py
```diff
@@ -288,6 +288,42 @@ def list_concepts(self, page_no: int = None,
       concept_info['id'] = concept_info.pop('concept_id')
       yield Concept(**concept_info)
 
+  def list_concept_relations(self,
+                             concept_id: str = None,
+                             predicate: str = None,
```
mogith-pn, line 293: Can we set predicate to optional, passing `None` to the request data is not the better way to do so. We can pass empty string in predicate which will work but not when directly given as `None`, 
When empty string is passed in predicate it will list all the relations with respect to that concept.

mogith-pn, line 293:
```suggestion
                             predicate: str = '',
```

mogith-pn (changes requested):
Made some comments, please test that changes locally, everything else is good.

clarifai/client/app.py
```diff
@@ -288,6 +288,42 @@ def list_concepts(self, page_no: int = None,
...
+                             page_no: int = None,
+                             per_page: int = None) -> Generator[ConceptRelation, None, None]:
+    """List all the concept relations of the app.
+
+    Args:
+        concept_id (str): The concept ID to filter the concept relations.
+        predicate (str): Type of relation to filter the concept relations.
```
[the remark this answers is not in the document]
srikanthbachala20: Done

clarifai/client/app.py
srikanthbachala20: Both None or empty string are working to get all the concept relations of a given concept

mogith-pn: Okay, then we can just make it optional, letting the users know that by default it returns all the relations.

mogith-pn: In this case add logging, if the request fails, We should return the error logging message in case user gives a different predicate in the request.

mogith-pn (changes requested):

clarifai/client/app.py
srikanthbachala20: Logging already present in the lister class

mogith-pn: Great then everything looks fine.

clarifai/client/app.py
```diff
@@ -288,6 +290,87 @@ def list_concepts(self, page_no: int = None,
...
+    """Print all the concept relations of the app in rich tree format.
+
+    Args:
+        relations_dict (dict): A dict of concept relations info.
+    """
+    for parent, children in relations_dict.items():
+      tree = Tree(parent)
+      for child in children:
+        tree.add(child)
+      rprint(tree)
```
sanjaychelliah, line 303: Can we move this function to logging utils, https://github.com/Clarifai/clarifai-python/blob/master/clarifai/utils/logging.py

clarifai/client/app.py
```diff
@@ -288,6 +290,87 @@ def list_concepts(self, page_no: int = None,
...
+        relations_dict[object_concept] = [subject_concept]
+    elif predicate == 'hypernym':
+      if subject_concept in relations_dict:
+        relations_dict[subject_concept].append(object_concept)
+      else:
+        relations_dict[subject_concept] = [object_concept]
+    else:
+      relations_dict[object_concept] = []
+      relations_dict[subject_concept] = []
+    return relations_dict
```
sanjaychelliah, line 325: Can we move this function also to utils folder?

sanjaychelliah (commented):
Added some comments, Otherwise LGTM!

clarifai/client/app.py
srikanthbachala20: Done

clarifai/client/app.py
srikanthbachala20: Moved to misc.py in utils folder
Done

sanjaychelliah (approved):
👍

PR #373, opened 2024-08-23
[DEVX-609]: Model Export Failures Bug Resolution

srikanthbachala20:
### Why

* Authentions headers missing, so unable to download the file

### How

* Added authentication headers

### DIFF
--- a/clarifai/client/model.py
+++ b/clarifai/client/model.py
@@ -11,6 +11,7 @@ from clarifai_grpc.grpc.api.resources_pb2 import Input
 from clarifai_grpc.grpc.api.status import status_code_pb2
 from google.protobuf.json_format import MessageToDict
 from google.protobuf.struct_pb2 import Struct, Value
+from requests.adapters import HTTPAdapter, Retry
 from tqdm import tqdm
 
 from clarifai.client.base import BaseClient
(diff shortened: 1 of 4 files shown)


sanjaychelliah (commented):
LGTM

clarifai/client/model.py
```diff
@@ -1295,15 +1296,22 @@ def _download_exported_model(
...
+      response = session.get(model_export_url, stream=True)
       response.raise_for_status()
 
       with open(local_filepath, 'wb') as f:
         progress = tqdm(
             total=model_export_file_size, unit='B', unit_scale=True, desc="Exporting model")
         for chunk in response.iter_content(chunk_size=8192):
-          f.write(chunk)
-          progress.update(len(chunk))
+          if chunk:
```
nitinbhojwani, line 1310: what is the intention behind this if else block?
We are not expecting chunk to be empty or None ever. We can undo this change.

nitinbhojwani (changes requested):

clarifai/client/model.py
srikanthbachala20: Intially, i thought there is some issue with the chucks, so I have added this if else block.I will undo this change now.

srikanthbachala20: Done

sanjaychelliah (approved):

nitinbhojwani (approved):

PR #375, opened 2024-08-27
[DEVX-605]: Dataset Annotations Export Bug Resolution

srikanthbachala20:
### Why

* Exporting annotation data of class concept or bounding box info (priority to class concept) [any one only]

### How

* Exporting annotation data of whatever present data(concept as well as bbox)


sanjaychelliah (approved):
👍

PR #385, opened 2024-09-12
[DEVX-641]: Model Export using Range Header

srikanthbachala20:
### Why

* Model export function is timing out from server-side when exporting on low-speed devices
* Ref : https://clarifai.atlassian.net/browse/DEVX-641

### How

* Used **Range** header in requests to download the model in chunks

### Tests

* To be added in test coverage ticket (https://clarifai.atlassian.net/browse/DEVX-596)


clarifai/client/model.py
```diff
@@ -1296,20 +1297,52 @@ def _download_exported_model(
...
-
       with open(local_filepath, 'wb') as f:
         progress = tqdm(
             total=model_export_file_size, unit='B', unit_scale=True, desc="Exporting model")
-        for chunk in response.iter_content(chunk_size=8192):
-          f.write(chunk)
-          progress.update(len(chunk))
-        progress.close()
+        downloaded_size = 0
+        range_size = 31457280  # 30MB
```
nitinbhojwani, line 1304: these variables like `range_size` and `chunk_size` are configurable so they can be read from a `constants` module

sanjaychelliah (approved):

sanjaychelliah (approved):
LGTM, except for the points raised by Nitin.

clarifai/client/model.py
srikanthbachala20: Done

nitinbhojwani (approved):

PR #388, opened 2024-09-19
Echo env vars

nishantasarma:
Used to echo env vars to be used in secrets

### Why
some information needed to setup secrets in another repo
*

### How

*

### Tests

*

### Notes

*


nishantasarma:
Got the required info.

PR #390, opened 2024-09-20
[EAGLE-4744] Added Hf loader and Handled Concepts in model upload

luv-bansal:
### What 

- Added HuggingFaceLoader in utils
- Added methods in model_upload for download_checkpoints and handling concepts for classification, detection models
- Changed `--model_path` flat in `model_upload.py` inplace of `-folder`

### Why

*

### How

*

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/runners/dockerfile_template/Dockerfile.cuda.template
+++ b/clarifai/runners/dockerfile_template/Dockerfile.cuda.template
@@ -49,37 +49,6 @@ COPY requirements.txt .
 # Install requirements and cleanup before leaving this line.
 # Note(zeiler): this could be in a future template as {{model_python_deps}}
 RUN python -m pip install -r requirements.txt && rm -rf /root/.cache
-# Install huggingface_hub with hf_transfer for fast download checkpoints
-RUN python -m pip install huggingface_hub[hf_transfer]
-#############################
-
-
-#############################
-# Download checkpoints
-#############################
-
-
-# Define HF_TOKEN build-time variables
-ARG HF_TOKEN
-
-# Set the HF_TOKEN environment variable
-ENV HF_TOKEN=${HF_TOKEN}
-
-# Copy the checkpoints directory into the container
-COPY 1/checkpoint[s] /app/checkpoints
-
-# Create checkpoint directory if it doesn't exist
-RUN mkdir -p /app/checkpoints
-
-# set permissions for the directory
-RUN chmod 777 /app/checkpoints
-
-# Note(luv): Uncomment the following lines for downloading checkpoints at build time.
-# Copy the download_checkpoints.py script into the container then run it to download checkpoints at build time.
-# COPY 1/download_checkpoints.py /app/download_checkpoints.py
-# RUN python /app/download_checkpoints.py
-############################
-
 
 #############################
 # Finally copy everything we built into a distroless image for runtime.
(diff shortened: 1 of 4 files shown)


clarifai/runners/dockerfile_template/Dockerfile.cuda.template
```diff
@@ -149,14 +118,11 @@ WORKDIR /app
 COPY . /app/model_dir/${name}
 
 # Copy the checkpoints directory into the distroless image.
-COPY --from=build /app/checkpoints /app/model_dir/${name}/1/checkpoints
```
zeiler, line 152: it wouldn't be from=build right?

clarifai/runners/models/model_upload.py
```diff
@@ -11,9 +11,15 @@
 
 from clarifai.client import BaseClient
 
+from clarifai.runners.utils.loader import HuggingFaceLoarder
+
 
 class ModelUploader:
   DEFAULT_PYTHON_VERSION = 3.11
+  CONCEPTS_REQUIRED_MODEL_TYPE = [
+      'visual-classifier', 'visual-detector', 'visual-segmenter', 'visual-embedder',
```
zeiler, line 20: the embedders shouldn't need concepts.

clarifai/runners/models/model_upload.py
```diff
@@ -212,13 +273,15 @@ def monitor_model_build(self, model_version_id):
...
   uploader = ModelUploader(folder)
+  uploader.download_checkpoints()
   uploader.create_dockerfile()
   uploader.upload_model_version()
 
 
 if __name__ == "__main__":
   parser = argparse.ArgumentParser()
-  parser.add_argument('--folder', type=str, help='Folder to tar and upload', required=True)
+  parser.add_argument(
```
zeiler, line 283: https://github.com/Clarifai/clarifai-python/tree/master/clarifai/runners/models i wonder if we should add a README.md to that folder to describe how to upload a model? Or should we just add that once you have the examples and we point to that repo in the README?

zeiler (approved):
just the dockerfile COPY from then you can merge.

clarifai/runners/models/model_upload.py
luv-bansal: Yeah, I think it would be better once we have examples in the `clarifai/examples`

clarifai/runners/dockerfile_template/Dockerfile.cuda.template
luv-bansal: oh right, just corrected it

luv-bansal: I just realized that we don't need to add the COPY checkpoints line in the Dockerfile, since we're already using `COPY . /app/model_dir/${name}` a line before. That line will also copy the checkpoint directory into the final folder of the container. What do you think?

luv-bansal:
Added some more changes:
- Added Clarifai dependency in Dockerfile, before it was missing 
- Removed COPY checkpoints line in the Dockerfile, since we're already using `COPY . /app/model_dir/${name} `a line before. That line will also copy the checkpoint directory into the final folder of the container
- Improve handling of importing `huggingface_hub` in HF loader

clarifai/runners/dockerfile_template/Dockerfile.cuda.template
zeiler: we might want to separate out the code from the checkpoint though in the future as copying in the code after the checkpoint would re-use most layers if you'd iterating on the same model code a few times hopefully it can re-use the checkpoint layer.

zeiler (approved):

zeiler (approved):

PR #391, opened 2024-09-23
Delete model_serving folder

luv-bansal:
### What 

- Deleted `model_serving` in this SDK, after the [Runners PR](https://github.com/Clarifai/clarifai-python/pull/389) has been merged, which introduced the new model serving method
- Deleted model serving tests

### Why

*

### How

*

### Tests

*

### Notes

*


clarifai/models/api.py
```diff
@@ -1,283 +0,0 @@
-# Copyright 2023 Clarifai, Inc.
```
zeiler, line 1: i don't think this file or the __init__.py was supposed to be deleted.

zeiler (changes requested):

clarifai/models/api.py
luv-bansal: Added this one back

zeiler (approved):

PR #392, opened 2024-09-24
[EAGLE-4201]: json logging

zeiler:
Added a json logger so it's convenient to get logs into logging stacks. Logs will look like: 

```
{"msg": "Loop iteration: 0 for thread 140485041844224", "@timestamp": "2024-09-24T21:31:15.411793Z", "filename": "test.py", "stack_info": null, "lineno": 214, "level": "info", "req_id": "runner-084a853704604c948bf226911f3cb0c6", "orig_req_id": "runner-084a853704604c948bf226911f3cb0c6"}
```
when env var `ENABLE_JSON_LOGGER=true`

### DIFF
--- a/clarifai/utils/logging.py
+++ b/clarifai/utils/logging.py
@@ -300,14 +300,14 @@ class JsonFormatter(logging.Formatter):
     if level:
       fields['level'] = level.lower()
 
-    # Get the thread local data from gRPC (proto/utils/common.py)
+    # Get the thread local data
     req_id = getattr(thread_log_info, 'req_id', None)
     if req_id:
       fields['req_id'] = req_id
     orig_req_id = getattr(thread_log_info, 'orig_req_id', None)
     if orig_req_id:
       fields['orig_req_id'] = orig_req_id
-    # Get the thread local data from gRPC (proto/utils/common.py)
+    # Get the thread local data
     requester = getattr(thread_log_info, 'requester', None)
     if requester:
       fields['requester'] = requester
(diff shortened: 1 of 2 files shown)


clarifai/utils/logging.py
```diff
@@ -151,3 +201,153 @@ def display_concept_relations_tree(relations_dict: Dict[str, Any]) -> None:
...
       tree.add(child)
     rprint(tree)
+
+
+def _default_json_default(obj):
+  """
+  Handle objects that could not be serialized to JSON automatically.
+
+  Coerce everything to strings.
+  All objects representing time get output as ISO8601.
```
wemoveon2, line 211: this in UTC?

wemoveon2 (approved):
tests?

clarifai/utils/logging.py
zeiler: yeah i think so. `2024-09-24T21:31:15.411793Z`

tests/test_misc.py
```diff
@@ -17,3 +19,49 @@ def test_get_logger_defaults():
   assert logger.level == logging.NOTSET
   assert logger.name == _get_library_name()
   assert isinstance(logger.handlers[0], RichHandler)
+
+
+def test_json_logger():
```
wemoveon2, line 24: did we not use pytest?

wemoveon2 (approved):

tests/test_misc.py
zeiler: pytest picks these up.

PR #403, opened 2024-10-03
improve logging and checkpoint download

zeiler:
Fixed issue comparing files in checkpoints. 
Unified all the logging to make it simpler and use an optional env var LOG_LEVEL to define what level to log at.

### DIFF
--- a/clarifai/utils/logging.py
+++ b/clarifai/utils/logging.py
@@ -19,6 +19,8 @@ from rich.tree import Tree
 
 install()
 
+# The default logger to use throughout the SDK is defined at bottom of this file.
+
 # For the json logger.
 JSON_LOGGER_NAME = "clarifai-json"
 JSON_LOG_KEY = 'msg'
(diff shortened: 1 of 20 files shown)


clarifai/utils/logging.py
```diff
@@ -357,3 +359,7 @@ def format(self, record):
           default=self.json_default,
           cls=self.json_cls,
       )
+
+
+# the default logger for the SDK.
+logger = get_logger(logger_level=os.environ.get("LOG_LEVEL", "INFO"), name="clarifai")
```
nitinbhojwani, line 365: Will the user be able to see the module logging the log line e.g. `compute_cluster` etc ?

nitinbhojwani (approved):

clarifai/utils/logging.py
zeiler: we'll see the filename and line number still.

PR #406, opened 2024-10-04
[EAGLE-5016]: New dockerfile

zeiler:
This updates dockerfiles to use debian and simplifies them. 

Added a flag to model upload to skip generating the dockerfile from template. This is helpful when debugging docker build or runtime issues as you can set that flag after manually editting the generated dockerfile. 

Also skip packaging up checkpoints unless --download_checkpoints is on during upload so that re-using folders for multiple uploads is easier.


clarifai/runners/dockerfile_template/Dockerfile.cuda.template
```diff
@@ -7,36 +7,21 @@
...
-    # drivers and nvidia-smi
-    nvidia-utils-${DRIVER_VERSION} \
-    nvidia-driver-${DRIVER_VERSION} \
-    libcap2-bin && \
+    libpython3-dev && \
     python${PYTHON_VERSION} -m venv /venv && \
     /venv/bin/pip install --disable-pip-version-check --upgrade pip setuptools wheel && \
     # Create a non-root user with minimal privileges and set file permissions
-    ln -sf /usr/bin/python${PYTHON_VERSION} /usr/bin/python3 && \
+    ln -sf /usr/bin/python3.10 /usr/bin/python3 && \
```
wemoveon2, line 22: fixed ver?

clarifai/runners/dockerfile_template/Dockerfile.cuda.template
```diff
@@ -7,36 +7,21 @@
 # * Create a non-root user with minimal privileges and use it
 ARG TARGET_PLATFORM=linux/amd64
```
wemoveon2, line 8: should use `TARGETPLATFORM` which gets set by docker instead

`ARG` also need to be after `FROM`

clarifai/runners/dockerfile_template/Dockerfile.cuda.template
```diff
@@ -57,54 +42,28 @@ RUN python -m pip install clarifai
...
+FROM --platform=$TARGET_PLATFORM gcr.io/distroless/python3-debian12:latest
+# FROM --platform=$TARGET_PLATFORM gcr.io/distroless/python3-debian12:debug
 ARG PYTHON_VERSION=${PYTHON_VERSION}
 # needed to call pip directly
 COPY --from=build /bin/sh /bin/sh
 
 # Copy driver libraries based on architecture
-# Set FOLDER_NAME based on TARGET_PLATFORM
 ENV TARGET_PLATFORM=${TARGET_PLATFORM}
 ENV FOLDER_NAME="x86_64-linux-gnu"
```
wemoveon2, line 53: should use `*-linux-gnu`

wemoveon2: *where globbing is possible like for `COPY`

clarifai/runners/dockerfile_template/Dockerfile.cuda.template
wemoveon2, line 1: missing all the symlink libs

wemoveon2:
```docker
FROM gcr.io/distroless/python3-debian12
ARG PYTHON_VERSION
# files for user profiles
COPY --from=build /etc/passwd /etc/passwd
COPY --from=build /etc/group /etc/group
# needed to call pip directly
COPY --from=build --chown=appuser:appuser /bin/bash /bin/sh
# virtual env
COPY --from=build --chown=appuser:appuser /venv /venv
# we have to overwrite the python3 binary that the distroless image uses
COPY --from=build /usr/bin/python${PYTHON_VERSION} /usr/bin/python3
# also copy in all the lib files for it.
COPY --from=build /usr/lib/python${PYTHON_VERSION} /usr/lib/python${PYTHON_VERSION}
# ld.so is needed for anything using shared libraries, automounted CUDA packages won't work without it
COPY --from=build /lib/*-linux-gnu/ld-linux-*.so.2 /lib/
COPY --from=build /lib/*-linux-gnu/libc.so.6 /lib/
COPY --from=build /lib/*-linux-gnu/libdl.so.2 /lib/
COPY --from=build /lib/*-linux-gnu/libpthread.so.0 /lib/
COPY --from=build /lib/*-linux-gnu/librt.so.1 /lib/
# ldconfig is needed to update the shared library cache so system libraries (like CUDA) can be found
COPY --from=build /usr/sbin/ldconfig /sbin/ldconfig
COPY --from=build /etc/ld.so.conf /etc/ld.so.conf
COPY --from=build /etc/ld.so.cache /etc/ld.so.cache
# ln is needed to create symbolic links (needed by nvidia-container-runtime)
COPY --from=build /usr/bin/ln /usr/bin/ln
# Set environment variables to use virtualenv by default
ENV VIRTUAL_ENV=/venv
ENV PYTHONPATH=${PYTHONPATH}:${VIRTUAL_ENV}/lib/python3.11/site-packages
ENV PATH=${VIRTUAL_ENV}/bin:/bin/bash:/bin/sh:${PATH}
# Switch to the non-root user
USER appuser
SHELL ["/bin/bash", "-c"]
RUN pip install torch
```

clarifai/runners/dockerfile_template/Dockerfile.cuda.template
zeiler: it seems to work when it's before. but maybe it's because we're templating it it just "works"?

zeiler: seems like ARG can be before FROM in their docs. I think we need the target platform to be set indendent of where the build is happening so we can build for other platforms.

clarifai/runners/dockerfile_template/Dockerfile.cuda.template
zeiler: oh removed this section as those vars aren't used.

clarifai/runners/dockerfile_template/Dockerfile.cuda.template
zeiler: yeah i'll leave that part to you after this one merges. That would be the second part of trying to get pip installs later. 

This does run already as is without copying over these libraries. I think the cuda stuff gets mounted in known places to torch automatically so we shouldn't have to copy lots of these.

clarifai/runners/dockerfile_template/Dockerfile.cuda.template
```diff
@@ -7,36 +7,21 @@
...
-    python${PYTHON_VERSION}-venv \
-    python${PYTHON_VERSION}-dev \
+    software-properties-common  \
     gcc \
-    libpython3-dev \
-    # drivers and nvidia-smi
-    nvidia-utils-${DRIVER_VERSION} \
-    nvidia-driver-${DRIVER_VERSION} \
-    libcap2-bin && \
+    libpython3-dev && \
```
zeiler, line 18: @wemoveon2 do we need this you think? I'm worried it'll instal some different version of python from what is in the base image.

wemoveon2: its the header files needed to build Python extensions. i don't think it will affect installed ver

zeiler: great

sainivedh (approved):
Just approving this to unblock.

PR #408, opened 2024-10-07
Improve experience for cv models

luv-bansal:
### What

- If no concepts are defined in the YAML file and the `download_checkpoints` flag is set to false, the user is prompted to download the checkpoints to load the concepts. Once the concepts are defined in config.yaml, the process continues with creating the tar file, but without including the checkpoints
### Why

*

### How

*

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/runners/models/model_upload.py
+++ b/clarifai/runners/models/model_upload.py
@@ -258,6 +258,18 @@ class ModelUploader:
     file_path = f"{self.folder}.tar.gz"
     logger.info(f"Will tar it into file: {file_path}")
 
+    model_type_id = self.config.get('model').get('model_type_id')
+
+    if not download_checkpoints and not os.path.exists(self.checkpoint_path) and (
+        model_type_id in self.CONCEPTS_REQUIRED_MODEL_TYPE) and 'concepts' not in self.config:
+      logger.error(
+          f"Model type {model_type_id} requires concepts to be specified in the config file or download the model checkpoints to infer the concepts."
+      )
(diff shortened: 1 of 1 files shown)


clarifai/runners/models/model_upload.py
```diff
@@ -258,6 +258,18 @@ def upload_model_version(self, download_checkpoints):
     file_path = f"{self.folder}.tar.gz"
     logger.info(f"Will tar it into file: {file_path}")
 
+    model_type_id = self.config.get('model').get('model_type_id')
+
+    if not download_checkpoints and not os.path.exists(self.checkpoint_path) and (
```
zeiler, line 263: instead of just exists(checkpoint_path) should it use the checkpoint loader validate function to make sure it's really a properly loaded checkpoint directory?

luv-bansal: yeah, that would be better just commit the update

clarifai/runners/models/model_upload.py
```diff
@@ -242,8 +247,7 @@ def _get_model_version_proto(self):
     model_type_id = self.config.get('model').get('model_type_id')
     if model_type_id in self.CONCEPTS_REQUIRED_MODEL_TYPE:
 
-      loader = HuggingFaceLoarder()
-      labels = loader.fetch_labels(self.checkpoint_path)
+      labels = HuggingFaceLoarder.fetch_labels(self.checkpoint_path)
```
zeiler, line 250: do we have a typo in Loader "Loarder"?

zeiler (approved):

PR #411, opened 2024-10-08
Runners Model run locally

luv-bansal:
### What
- This PR implements a `ModelRunLocally` class that allows users to test and run a runners model locally. 
- It creates a virtual environment, installs required dependencies from `requirements.txt`, and and tests the model locally.
- Anyone can simply test their model using
```bash
python -m clarifai.runners.models.model_run_locally --model_path <model_path>
```

### Why
- It simplify the process of testing models locally. 
-  By using a virtual environment, it prevents dependency conflicts in requierements.txt.
- Prevent from model upload failure due to typos in model.py or wrong implementation of model

### Tests
- Tested by running the models examples locally
### Notes
- Not sure if `model_run_locally` is the best name, let me know if need to change

### DIFF
--- a/clarifai/runners/models/model_run_locally.py
+++ b/clarifai/runners/models/model_run_locally.py
@@ -23,6 +23,7 @@ class ModelRunLocally:
     self.model_path = model_path
     self.requirements_file = os.path.join(self.model_path, "requirements.txt")
     self.venv_dir, self.temp_dir = self.create_temp_venv()
+    self.python_executable = os.path.join(self.venv_dir, "bin", "python")
 
   def create_temp_venv(self):
     """Create a temporary virtual environment."""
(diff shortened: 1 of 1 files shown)


clarifai/runners/models/model_run_locally.py
```diff
@@ -0,0 +1,192 @@
...
+    try:
+      logger.info("Testing the model locally...")
+      subprocess.check_call(command)
+      logger.info("Model tested successfully!")
+    except subprocess.CalledProcessError as e:
+      logger.error(f"Error testing the model: {e}")
+      sys.exit(1)
+
+  # run the model server
+  def run_model_server(self):
```
luv-bansal, line 152: I include this method to start the model runner server locally, but not sure if's useful here

zeiler (approved):
This is great! I think we should have examples of how to call the model very easily too and maybe print out when the local dev server starts how to call it.

PR #414, opened 2024-10-09
Make `input_type` as Optional in Model/Workflow Predict 

sainivedh:
### Why

* Making `input_type` as optional during Model/Workflow for better User experience.
```python
>>> from clarifai.client.model import Model
>>> model = Model("https://clarifai.com/openai/chat-completion/models/GPT-4")
>>> stream_response = model.generate_by_bytes(b'Write a tweet on future of AI',
                                                      deployment_id='deployment_id',
                                                      inference_params=dict(temperature=str(0.7), max_tokens=30)))
```

### How

* Get Model `input_type` internally with `GetModelType` endpoint
* Endpoint is cached, so even for first predict extra latency should be negligible

### Tests

* Modified existing `Models/Workflow` tests

### Notes

* To support backward compatibility this not entirely removed as we have SDK predict code everywhere through out (examples, model notes, docs...etc)


README.md
```diff
@@ -223,19 +223,19 @@ print(gpt_4_model)
 
 
 # Model Predict
-model_prediction = Model("https://clarifai.com/anthropic/completion/models/claude-v2").predict_by_bytes(b"Write a tweet on future of AI", input_type="text")
+model_prediction = Model("https://clarifai.com/anthropic/completion/models/claude-v2").predict_by_bytes(b"Write a tweet on future of AI")
```
zeiler, line 226: love this PR. I think you should still show an example of setting input_type though as well because for a model that takes in "Any" to "Any" you have to specify (or use the protos in predict(), generate(), stream()) is maybe a better recommendation?

zeiler (commented):
awesome, just left a comment to show some examples for Any input type.

clarifai/client/model.py
```diff
@@ -802,9 +850,17 @@ def generate_by_url(self,
...
         >>> list_stream_response = [response for response in stream_response]
     """
+    if not input_type:
+      model_input_types = self.get_model_input_types()
+      if len(model_input_types) > 1:
+        raise UserError(
+            "Model has multiple input types. Please use model.predict() for this multi-modal model."
+        )
+      input_type = model_input_types[0]
+
```
sanjaychelliah, line 863: I would suggest may be having this in a internal function(`def _check_input_type()` or something) and reuse that since this was used in multiple functions.

sanjaychelliah (commented):
Great Addition 👏 , added a comment!

clarifai/client/model.py
sainivedh: Done

README.md
sainivedh: > or use the protos in predict(), generate(), stream()) is maybe a better recommendation?

Correct.

sanjaychelliah (approved):
LGTM!

PR #416, opened 2024-10-09
fixing streamlit query_params call to remove error w/streamlit >=1.30

gglickstein:
New versions of streamlit lead to error text boxes when using the removed st.experimental_get_query_params() call. It has been replaced by st.query_params which is not callable, so it's passed to a dict to replicate the functionality.

### DIFF
--- a/clarifai/client/auth/helper.py
+++ b/clarifai/client/auth/helper.py
@@ -136,7 +136,7 @@ class ClarifaiAuthHelper:
 
     # Then add in the query params.
     try:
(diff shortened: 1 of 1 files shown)


mogith-pn:
Looks good ! 
Although to hide errors or supress warnings from the main UI, we can use something like this to set the configuration in the config.toml file.
https://docs.streamlit.io/develop/api-reference/configuration/config.toml#view-all-configuration-options

mogith-pn (approved):
LGTM.

mansi-k (approved):

PR #418, opened 2024-10-09
[EAGLE-4760] allow no checkpoint to fail gracefully

wemoveon2:
### Why

* If no checkpoint is specified we end up throwing a `TypeError` due to trying to unpack `None`

### How

* Refactor validation logic to return if no checkpoint provided

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/runners/models/model_upload.py
+++ b/clarifai/runners/models/model_upload.py
@@ -1,8 +1,8 @@
 import argparse
 import os
 import time
-from string import Template
 import typing
+from string import Template
 
 import yaml
 from clarifai_grpc.grpc.api import resources_pb2, service_pb2
@@ -200,7 +200,7 @@ class ModelUploader:
   def download_checkpoints(self) -> None:
     if not self.config.get("checkpoints"):
       logger.info("No checkpoints specified in the config file")
-      return 
+      return
     repo_id, hf_token = self._validate_config_checkpoints()
 
     loader = HuggingFaceLoader(repo_id=repo_id, token=hf_token)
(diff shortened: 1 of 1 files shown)


clarifai/runners/models/model_upload.py
```diff
@@ -210,7 +210,7 @@ def download_checkpoints(self):
       logger.error(f"Failed to download checkpoints for model {repo_id}")
     else:
       logger.info(f"Downloaded checkpoints for model {repo_id}")
-    return success
+    return
```
luv-bansal, line 213: we want to return success or not so that we can use it when trying to do here https://github.com/Clarifai/models-images/pull/29/files to fail the init build if it’s not successfull. we return None and assert.

clarifai/runners/models/model_upload.py
```diff
@@ -59,11 +60,7 @@ def _load_config(config_file: str):
       config = yaml.safe_load(file)
     return config
 
-  def _validate_config_checkpoints(self):
-    if not self.config.get("checkpoints"):
-      logger.info("No checkpoints specified in the config file")
-      return None
-
+  def _validate_config_checkpoints(self) -> typing.Tuple[str, str]:
```
luv-bansal, line 63: I also created PR to takes care of this
https://github.com/Clarifai/clarifai-python/pull/417/files#diff-7ee953fca56b92c0015d093416f9a390d19331e5cea0e845f85dfddb6cef4117R65

clarifai/runners/models/model_upload.py
wemoveon2: what do you mean? the PR you linked made it so `success` isn't used?

clarifai/runners/models/model_upload.py
wemoveon2: ok go ahead and merge that then.

PR #420, opened 2024-10-10
Improve UX for model upload

luv-bansal:
### Why

* Only download HF config.json file instead of downloading complete model checkpoints to infer the concepts
* Improving logging
* Fixed model-upload bug when checkpoints not specified in config.yaml

### How

*

### Tests

* Tested Manually

### Notes

*


zeiler (approved):

sainivedh:
@luv-bansal I've fixed the failing runner tests. We don't allow passing of `RunnerID` anymore. It'll always be a `UUID`

PR #423, opened 2024-10-14
Bump llama version

sanjaychelliah:
### Why

* Security update: Bumping llama index version

### How

* https://github.com/Clarifai/clarifai-python/security/dependabot

### Tests

* Existing tests are sufficient


mogith-pn (approved):
LGTM.

PR #424, opened 2024-10-15
[DEVX-572]: Removing 'id-' prefix from concept ids from SDK

srikanthbachala20:
### Why

* To remove unnecessary prefixes to concept ID added from SDK

### How

* Remove the 'id-' prefix added to concept ID using SDK

### Tests

* Already present


sanjaychelliah (approved):
LGTM

PR #426, opened 2024-10-17
Fix tests

luv-bansal:
### What

- Removed stream tests for now, and will add again when stream API will be fix
- Fix app test which was failing before

### Why

*

### How

*

### Tests

*

### Notes

*

### DIFF
--- a/tests/runners/test_runners.py
+++ b/tests/runners/test_runners.py
@@ -220,17 +220,18 @@ class TestRunnerServer:
     for i, res in enumerate(stub.GenerateModelOutputs(req)):
       self._validate_response(res, text + out.format(i=i))
 
-  # def test_stream(self):
-  #   text = "This is a long text for testing stream"
-  #   out = "Stream Hello World {i}"
-  #   req = self._format_request(text)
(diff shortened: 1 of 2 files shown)


tests/runners/test_runners.py
```diff
@@ -220,17 +220,17 @@ def test_generate(self):
     for i, res in enumerate(stub.GenerateModelOutputs(req)):
       self._validate_response(res, text + out.format(i=i))
 
-  def test_stream(self):
-    text = "This is a long text for testing stream"
-    out = "Stream Hello World {i}"
-    req = self._format_request(text)
+  # def test_stream(self):
```
sanjaychelliah, line 223: Rather than commenting out, can we add skip with a TODO comment like [this one](https://github.com/Clarifai/clarifai-python/blob/2aaf180badafba9dc06218688de53bffad378ec8/tests/compute_orchestration/test_crud_operations.py#L107)?

luv-bansal: oh this is nice, let me update it

sanjaychelliah (approved):
👍

PR #430, opened 2024-10-25
[DEVX-738]: Request ID Prefix Header Addition

srikanthbachala20:
### Why

* To monitor the SDK usage

### How

* Added a request-id-prefix header to SDK requests to differentiate them from other requests

### DIFF
--- a/clarifai/client/auth/helper.py
+++ b/clarifai/client/auth/helper.py
@@ -6,9 +6,12 @@ from urllib.parse import urlparse
 from clarifai_grpc.channel.clarifai_channel import ClarifaiChannel
 from clarifai_grpc.grpc.api import resources_pb2, service_pb2_grpc
 from clarifai.utils.constants import CLARIFAI_PAT_ENV_VAR, CLARIFAI_SESSION_TOKEN_ENV_VAR
+from clarifai import __version__
 
 DEFAULT_BASE = "https://api.clarifai.com"
 DEFAULT_UI = "https://clarifai.com"
+
+REQUEST_ID_PREFIX = "sdk-python-" + str(__version__)
 
 # Map from base domain to True / False for whether the base has https or http.
 # This is filled in get_stub() if it's not in there already.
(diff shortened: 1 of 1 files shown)


clarifai/client/auth/helper.py
```diff
@@ -5,11 +5,15 @@
...
 from clarifai_grpc.channel.clarifai_channel import ClarifaiChannel
 from clarifai_grpc.grpc.api import resources_pb2, service_pb2_grpc
+
+from clarifai import __version__
 from clarifai.utils.constants import CLARIFAI_PAT_ENV_VAR, CLARIFAI_SESSION_TOKEN_ENV_VAR
 
 DEFAULT_BASE = "https://api.clarifai.com"
 DEFAULT_UI = "https://clarifai.com"
 
+REQUEST_ID_PREFIX = "sdk-python-" + str(__version__)
```
nitinbhojwani, line 15: Since `__version__` is already a str, it can be

```
REQUEST_ID_PREFIX = f"sdk-python-{__version__}"
```

clarifai/client/auth/helper.py
```diff
@@ -269,9 +273,11 @@ def metadata(self):
       metadata: the metadata need to send with all grpc API calls in the API client.
     """
     if self._pat != "":
-      return (("authorization", "Key %s" % self._pat),)
+      return (("authorization", "Key %s" % self._pat), ("x-clarifai-request-id-prefix",
```
nitinbhojwani, line 276: `x-clarifai-request-id-prefix` can also be a constant like `REQUEST_ID_PREFIX_HEADER`

nitinbhojwani (changes requested):
Suggested two small changes

clarifai/client/auth/helper.py
abhash-ai: Let me know if it needs to change

abhash-ai, line 15: How about GRPC case? or is that a different repo?

srikanthbachala20: Yes

srikanthbachala20: Done

clarifai/client/auth/helper.py
srikanthbachala20: Done

clarifai/client/auth/helper.py
nitinbhojwani: > Note: version must be in format: major.minor

@srikanthbachala20 did you test with full semantic version(`major.minor.patch`) and it works? i.e. Is this trimmed automatically by BE? 

If it doesn't work, please also trim it to `major.minor` instead of default `major.minor.patch`:

```
REQUEST_ID_PREFIX = f"sdk-python-{'.'.join(__version__.split('.')[:2])}"
```

srikanthbachala20: Yes, I have tested with full semantic version. It is trimmed automatically by BE

nitinbhojwani (approved):

PR #442, opened 2024-11-11
[DEVX-824]: Added Delete Annotations Feature

srikanthbachala20:
### Why

* To provide functionality to delete annotations as per user requirements

### How

* Added a function in **Input** class of `clarifai/client/input.py` to delete annotations using clarifai-grpc library

### Tests

* Added in `tests/test_data_upload.py` file


sanjaychelliah (approved):
LGTM. 
Great Addition. Can we quickly add this in example repo too!

PR #452, opened 2024-11-25
[EAGLE-4467][EAGLE-4497] Add streaming tests and passing inference parameters to runners tests

luv-bansal:
### What

- Added tests for all stream and generate methods by starting the local dev runner
- Added tests for passing inference parameters to runners

### Why

*

### How

*

### Tests

*

### Notes

*


zeiler (approved):
nice looks like you found a bug in generate too!

PR #456, opened 2024-12-02
Download data.parts as bytes in the API

luv-bansal:
### What

- Since we now have the `data.parts` proto, which supports multiple images and other inputs in a single request, we also need to downloading them as bytes in the API

### Why

*

### How

*

### Tests

*

### Notes

*


zeiler (approved):
nice. i'm now wondering if we'll need it on frames or regions? we can leave as is and do in future as needed.

PR #459, opened 2024-12-10
[DEVX-839]: CLI support for Model Predict

srikanthbachala20:
### Why

* For ease of model prediction using CLI

### How

* Adding predict command for model CLI submodule

### Tests

* Already exists

### DIFF
--- a/clarifai/cli/model.py
+++ b/clarifai/cli/model.py
@@ -112,22 +112,21 @@ def predict(ctx, config, model_id, user_id, app_id, model_url, file_path, url, b
   from clarifai.utils.cli import from_yaml
   if config:
     config = from_yaml(config)
-    model_id = config.get('model_id', model_id)
-    user_id = config.get('user_id', user_id)
-    app_id = config.get('app_id', app_id)
-    model_url = config.get('model_url', model_url)
-    file_path = config.get('file_path', file_path)
-    url = config.get('url', url)
-    bytes = config.get('bytes', bytes)
-    input_id = config.get('input_id', input_id)
-    input_type = config.get('input_type', input_type)
-    compute_cluster_id = config.get('compute_cluster_id', compute_cluster_id)
-    nodepool_id = config.get('nodepool_id', nodepool_id)
-    deployment_id = config.get('deployment_id', deployment_id)
-    inference_params = config.get('inference_params', inference_params)
-    output_config = config.get('output_config', output_config)
-  if sum([1 for opt in [model_id, model_url] if opt]) != 1:
-    raise ValueError("Exactly one of --model_id or --model_url must be provided.")
+    model_id, user_id, app_id, model_url, file_path, url, bytes, input_id, input_type, compute_cluster_id, nodepool_id, deployment_id, inference_params, output_config = (
+        config.get(k, v)
+        for k, v in [('model_id', model_id), ('user_id', user_id), ('app_id', app_id), (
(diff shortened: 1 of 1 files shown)


clarifai/cli/model.py
luv-bansal, line 1: Please add similar CLI methods for `generate` and `stream`

clarifai/cli/model.py
```diff
@@ -68,3 +69,130 @@ def run_locally(model_path):
...
+      runner_selector = Nodepool.get_runner_selector(
+          user_id=ctx.obj['user_id'],
+          compute_cluster_id=compute_cluster_id,
+          nodepool_id=nodepool_id)
+    model_prediction = model.predict(
+        inputs=inputs,
+        runner_selector=runner_selector,
+        inference_params=inference_params,
+        output_config=output_config)
+  click.echo(f"Prediction successful:\n{model_prediction}")
```
luv-bansal, line 198: There is might be the case where prediction failed, I think we can just print `model_prediction`
`click.echo(model_prediction)`

luv-bansal (commented):
Looks good!! 

Add similar CLI methods for `generate` and `stream`

clarifai/cli/model.py
srikanthbachala20: Please create a separate ticket for it

luv-bansal (approved):

clarifai/cli/model.py
```diff
@@ -68,3 +69,130 @@ def run_locally(model_path):
...
+    file_path = config.get('file_path', file_path)
+    url = config.get('url', url)
+    bytes = config.get('bytes', bytes)
+    input_id = config.get('input_id', input_id)
+    input_type = config.get('input_type', input_type)
+    compute_cluster_id = config.get('compute_cluster_id', compute_cluster_id)
+    nodepool_id = config.get('nodepool_id', nodepool_id)
+    deployment_id = config.get('deployment_id', deployment_id)
+    inference_params = config.get('inference_params', inference_params)
+    output_config = config.get('output_config', output_config)
```
sanjaychelliah, line 128: WDYT of this suggestion, this is redundant code, so maybe in a one line would be fine.

```suggestion
 model_id, user_id, app_id, model_url, file_path, url, bytes, input_id, input_type, compute_cluster_id, nodepool_id, deployment_id, inference_params, output_config = (config.get(k, v) for k, v in [('model_id', model_id), ('user_id', user_id), ('app_id', app_id), ('model_url', model_url), ('file_path', file_path), ('url', url), ('bytes', bytes), ('input_id', input_id), ('input_type', input_type), ('compute_cluster_id', compute_cluster_id), ('nodepool_id', nodepool_id), ('deployment_id', deployment_id), ('inference_params', inference_params), ('output_config', output_config)]))
```

clarifai/cli/model.py
```diff
@@ -68,3 +69,130 @@ def run_locally(model_path):
...
+  if sum([1 for opt in [model_id, model_url] if opt]) != 1:
```
sanjaychelliah, line 129: if `model_url` is provided, the `model_id`, `user_id`, `app_id` should not be entered. We should have a condition check for those.

clarifai/cli/model.py
sanjaychelliah: This is fine I think, If model prediction fails, [this](https://github.com/Clarifai/clarifai-python/blob/282238dc8aba7b25811c819c619d510c760f2d25/clarifai/client/model.py#L448) will take care of the failure message I believe.

sanjaychelliah (approved):

PR #473, opened 2025-01-08
Improved error logging

mogith-pn:
### Why

* When called from the upload_dataset with dataloader consists of image bytes the error is not catching properly. So added a failsafe checks and improved error logging.


srikanthbachala20 (approved):
LGTM

PR #479, opened 2025-01-14
[EAGLE-5408]: Add storage request inferred from tar and checkpoint size

patricklundquist:
### Why
* prevent build failures and poor resource allocation due to wrong storage requests. 

### How
*

### Tests
* clarifai model upload --model_path ../examples/models/model_upload/llms/llama-3-8b-instruct

### Notes
*


clarifai/runners/models/model_upload.py
```diff
@@ -536,6 +548,50 @@ def monitor_model_build(self):
...
+      int: The total size of the contents in bytes.
+    """
+    total_size = 0
+    with tarfile.open(tar_file_path, 'r') as tar:
+      for member in tar:
+        if member.isfile():
+          total_size += member.size
+    return total_size
+
+  def get_huggingface_checkpoint_total_size(self, repo_name):
```
luv-bansal, line 568: Seems like currently this function not works with restricted models, we need to use `hf_token`  for authentication. 

we could also add try except block too for better handling of error.
I think we could add this method in [HF loader class](https://github.com/Clarifai/clarifai-python/blob/master/clarifai/runners/utils/loader.py#L9) for better structure and that class has other method for validating model repo if it's exists, etc

luv-bansal (commented):
Looks good, but I have doubts about whether the `get_huggingface_checkpoint_total_size` function will work for restricted models.

If it’s working, we can merge it for now. In the follow-up PR, I’ll address the comment I added.

clarifai/runners/models/model_upload.py
patricklundquist: Done. Changes seem to work. 

Need to wait for clarifai-python-grpc to get updated and then bump up that version in requirements.txt

luv-bansal (approved):

PR #486, opened 2025-01-15
[EAGLE-5418] add timestamp tag to images

wemoveon2:
### Why

*

### How

*

### Tests

*

### Notes

*


luv-bansal (commented):
@wemoveon2 we already added `GIT_SHA` in the image tag, so I don't think we need this PR anymore, can we close this?

PR #488, opened 2025-01-17
[EAGLE-5416] Added Tests for Download checkpoints and Fix download methods

zeiler:
- Added download checkpoints tests without setting PAT
- remove `.cache` directory while downloading checkpoints
- Fixed `validate_download` method: it got broke before because of adding `ignore_patterns` while downloading checkpoints. 
- Handle `.` as model_path better
- Do less validation if just calling download_checkpoints. 
- Fall back to env var HF_TOKEN if not provided in the checkpoints section of config.yaml


ackizilkale (approved):
Great

PR #492, opened 2025-01-22
[EAGLE-5342] Added Tests for  Model run locally

luv-bansal:
### What

- Added tests for Model run locally within both virtual env and docker container
- Tests covers all the related function used in following command:
  - `clarifai model test-locally --model_path <> --model env`
  - `clarifai model test-locally --model_path <> --model container`
- I also fixed some edge cases for the `test-locally` and `run-locally` commands to ensure they work across all operating systems, and also works whether or not the user's system has a GPU.

### Why

*

### How

*

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/runners/models/model_run_locally.py
+++ b/clarifai/runners/models/model_run_locally.py
@@ -28,7 +28,7 @@ class ModelRunLocally:
     self.requirements_file = os.path.join(self.model_path, "requirements.txt")
 
     # ModelUploader contains multiple useful methods to interact with the model
-    self.uploader = ModelUploader(self.model_path, download_validation_only=True)
+    self.uploader = ModelUploader(self.model_path, validate_api_ids=False)
     self.config = self.uploader.config
 
   def _requirements_hash(self):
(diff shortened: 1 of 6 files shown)


clarifai/runners/models/model_run_locally.py
```diff
@@ -27,14 +28,31 @@ def __init__(self, model_path):
     self.requirements_file = os.path.join(self.model_path, "requirements.txt")
 
     # ModelUploader contains multiple useful methods to interact with the model
-    self.uploader = ModelUploader(self.model_path)
+    self.uploader = ModelUploader(self.model_path, validate_api_ids=False)
```
zeiler, line 31: i chagned the ModelUploader arg to use to be a new one called `download_validation_only` we should switch to only use that and remove the old arg. https://github.com/Clarifai/clarifai-python/blob/master/clarifai/runners/models/model_upload.py#L32

zeiler (approved):
nice, good improvements. Just that one arg should change or you can do that in a follow PR and make sure all places are changed.

PR #498, opened 2025-01-24
[EAGLE-5448]: upload improvements

zeiler:
Use new base images within the dockerfile. 
Fix checking for torch CPU version. 
Allow CLI to be run from python -m as needed.
Added a download checkpoints function to CLI.

This PR is expected to fail the tests as it needs the changes in here in the dockerfile.


clarifai/runners/dockerfile_template/Dockerfile.template
```diff
@@ -1,43 +1,82 @@
...
+    TORCHINDUCTOR_CACHE_DIR=/tmp/torchinductor_cache \
+    HOME=/tmp \
+    DEBIAN_FRONTEND=noninteractive
 
-# Set the working directory to /app
-WORKDIR /app
+#####
+# Copy the python requirements needed to download checkpoints
+#####
+COPY --link=true --from=downloader /venv /venv
```
wemoveon2, line 40: shouldnt this be from `pip_packages`

wemoveon2: ok it seems like we just need one COPY from `pip_packages` since all `downloader` does is reinstall the same packages as `pip_packages`

wemoveon2 (approved):

clarifai/runners/dockerfile_template/Dockerfile.template
zeiler: right this line just get the basic clarifai install but not all the stuff from the user's requirements.txt. just enough to download then after download we overwrite /venv again from pip_packages.

PR #500, opened 2025-01-28
[DEVX-896]: Update Model Predict CLI

srikanthbachala20:
### Why

* To support multi-input models predictions
* get_input usage is removed from Model Predict CLI as unable to provide the app_id (given app_id is of model)
* Added get_input method to support Model Predict CLI in future

### How

* Removed the exception part to guarantee single type of input
* Removed get_input usage in Model Predict CLI
* Added get_input method to Inputs class

### Tests

* Added in tests/test_data_upload.py

### DIFF
--- a/clarifai/cli/model.py
+++ b/clarifai/cli/model.py
@@ -173,8 +173,6 @@ def run_locally(model_path, port, mode, keep_env, keep_image):
 @click.option('--file_path', required=False, help='File path of file for the model to predict')
 @click.option('--url', required=False, help='URL to the file for the model to predict')
 @click.option('--bytes', required=False, help='Bytes to the file for the model to predict')
-@click.option(
-    '--input_id', required=False, help='Existing input id in the app for the model to predict')
 @click.option('--input_type', required=False, help='Type of input')
 @click.option(
     '-cc_id',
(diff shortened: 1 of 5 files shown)


luv-bansal (commented):
The Predict CLI is still broken due to the two types of user_id used in the predict methods. Specifically, if cc, np, or dpl is used with the predict CLI for predictions with a dedicated model, it will not work.

@sanjaychelliah I'm not sure if we want to address this in this PR or a follow-up PR. We can merge this, but a separate ticket should be created to fix above issue.

sanjaychelliah:
> The Predict CLI is still broken due to the two types of user_id used in the predict methods. Specifically, if cc, np, or dpl is used with the predict CLI for predictions with a dedicated model, it will not work.
> 
> @sanjaychelliah @srikanthbachala20 I'm not sure if we want to address this in this PR or a follow-up PR. We can merge this, but a separate ticket should be created to fix above issue.

Hey @luv-bansal . We'll do that in a separate PR. Thanks for reminding!

luv-bansal (approved):

sanjaychelliah (approved):

PR #511, opened 2025-02-06
[PR-306] PAT feedback

ackizilkale:
### Why

The logic is unable to enter logger error logic

### DIFF
--- a/clarifai/runners/models/model_run_locally.py
+++ b/clarifai/runners/models/model_run_locally.py
@@ -475,7 +475,7 @@ def main(model_path,
          keep_env=False,
          keep_image=False):
 
-  if not os.environ['CLARIFAI_PAT']:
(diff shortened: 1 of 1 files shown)


clarifai/runners/models/model_run_locally.py
```diff
@@ -475,7 +475,7 @@ def main(model_path,
          keep_env=False,
          keep_image=False):
 
-  if not os.environ['CLARIFAI_PAT']:
+  if 'CLARIFAI_PAT' not in os.environ:
```
zeiler, line 478: might want 
```
if not os.environ.get("CLARIFAI_PAT", None)
```
so that it validates empty PATs as incorrect as well

ackizilkale: yeah I guess that is also a possibility.

zeiler (approved):

PR #515, opened 2025-02-12
[EAGLE-5536]: 3 options to download checkpoints

zeiler:


luv-bansal:
@zeiler 
I feel that adding these lines in the `load_model` method just to retrieve the checkpoint path and download the checkpoint does not provide a good user experience in `runtime` stage
```
      model_path = os.path.dirname(os.path.dirname(__file__))
      builder = ModelBuilder(model_path, download_validation_only=True)
      checkpoints = builder.download_checkpoints(stage="runtime")

```
User have the option to directly provide any HF repo ID (e.g., meta-llama/Llama-3.1-8B-Instruct) in AutoModelForSeq2SeqLM.from_pretrained('meta-llama/Llama-3.1-8B-Instruct'), which also downloads the checkpoints at runtime in an optimized way, I think using our built-in function for downloading checkpoints at runtime in model.py could be confusing for users.

clarifai/cli/model.py
```diff
@@ -29,7 +29,11 @@ def model():
 def upload(model_path, download_checkpoints, skip_dockerfile):
   """Upload a model to Clarifai."""
   from clarifai.runners.models.model_builder import upload_model
-  upload_model(model_path, download_checkpoints, skip_dockerfile)
+  if download_checkpoints:
+    stage = "any"  # ignore config.yaml and force download now.
```
luv-bansal, line 33: stage `"any"` is confusion, can we limit it to `runtime`, `build` or `upload`

clarifai/cli/model.py
```diff
@@ -44,14 +48,23 @@ def upload(model_path, download_checkpoints, skip_dockerfile):
...
     help=
-    'Option path to write the checkpoints to. This will place them in {out_path}/ If not provided it will default to {model_path}/1/checkpoints where the config.yaml is read..'
+    'Option path to write the checkpoints to. This will place them in {out_path}/1/checkpoints If not provided it will default to {model_path}/1/checkpoints where the config.yaml is read.'
+)
+@click.option(
+    '--stage',
+    type=str,
+    required=False,
+    default="build",
+    show_default=True,
```
luv-bansal, line 58: here `type` could be `type=click.Choice(['runtime', 'build', 'upload'], case_sensitive=False)` which only allow user to choose among "runtime," "build," or "upload"

clarifai/runners/models/model_builder.py
```diff
@@ -702,10 +734,16 @@ def monitor_model_build(self):
...
 
 
-def upload_model(folder, download_checkpoints, skip_dockerfile):
+def upload_model(folder, stage, skip_dockerfile):
+  """
+  Uploads a model to Clarifai.
+
+  :param folder: The folder containing the model files.
+  :param stage: The stage of when you're uploading this model. This is used to determine when to download the checkpoints based on a match with the "when" field in the config.yaml checkpoints section or if you set stage to "any" it will always download the checkpoints.
+  :param skip_dockerfile: If True, skips creating the Dockerfile so you can re-use the local one.
```
luv-bansal, line 743: when stage is set to `any` then in which stage it will download checkpoint? 

`any` stage seems to be confusion, we could limit it to only runtime, build or upload which one of them default

tests/runners/test_download_checkpoints.py
```diff
@@ -58,5 +58,6 @@ def test_validate_download(checkpoint_dir):
 def test_download_checkpoints(dummy_runner_models_dir):
   model_folder_path = os.path.join(os.path.dirname(__file__), "dummy_runner_models")
   model_builder = ModelBuilder(model_folder_path, download_validation_only=True)
-  isdownloaded = model_builder.download_checkpoints()
-  assert isdownloaded is True
+  checkpoint_dir = model_builder.download_checkpoints(stage="any")  # any forces download now.
+  assert checkpoint_dir == os.path.join(
+      os.path.dirname(__file__), "dummy_runner_models", "1", "checkpoints")
```
luv-bansal, line 63: windows tests are failing here because of different path syntax in windows and we are hard coding `self.checkpoint_suffix` to be `1/checkpoints` 

```
>     assert checkpoint_dir == os.path.join(
          os.path.dirname(__file__), "dummy_runner_models", "1", "checkpoints")
E     AssertionError: assert 'D:\\a\\clari...1/checkpoints' == 'D:\\a\\clari...\\checkpoints'
E       Skipping 62 identical leading characters in diff, use -v to show
E       - r_models\1\checkpoints
E       ?           ^
E       + r_models\1/checkpoints
E       ?           
```

luv-bansal (approved):

PR #524, opened 2025-02-28
[PR-388] Add num_threads setting for serving

phatvo9:
Introduce `num_threads` in config.yaml enabling setting multiple threads serving. Previously can be set only by `CLARIFAI_NUM_THREADS` env variable. 

### Why

* Keep it in config.yaml is easier to config, maintain and debug.
* Improve RPM to handle concurrent requests more efficiently.
### How

The variable can be set through the environment or via the `num_threads` in config.yaml, with config.yaml taking priority over the environment variable.

### Tests

* add test case

### DIFF
--- a/clarifai/runners/models/model_builder.py
+++ b/clarifai/runners/models/model_builder.py
@@ -226,6 +226,15 @@ class ModelBuilder:
           )
           logger.info("Continuing without Hugging Face token")
 
+    num_threads = self.config.get("num_threads")
+    if num_threads or num_threads == 0:
+      assert isinstance(num_threads, int) and num_threads >= 1, ValueError(
+          f"`num_threads` must be an integer greater than or equal to 1. Received type {type(num_threads)} with value {num_threads}."
+      )
+    else:
+      num_threads = int(os.environ.get("CLARIFAI_NUM_THREADS", 1))
+      self.config["num_threads"] = num_threads
+
   @staticmethod
   def _get_tar_file_content_size(tar_file_path):
     """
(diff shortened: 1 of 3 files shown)


clarifai/runners/models/model_builder.py
```diff
@@ -522,11 +531,28 @@ def hf_labels_to_config(self, labels, config_file):
     logger.info(f"Updated config.yaml with {len(concepts)} concepts.")
 
   def get_model_version_proto(self):
-
+    from clarifai.client import Model
```
ackizilkale, line 534: These do not belong here right?

clarifai/runners/models/model_builder.py
```diff
@@ -522,11 +531,28 @@ def hf_labels_to_config(self, labels, config_file):
...
+        "field_type": 3,
+        "default_value": 0.7
+    }]
+    inference_param_proto = Model._make_inference_params_proto(inference_params)
+    print(inference_param_proto)
     model_version_proto = resources_pb2.ModelVersion(
         pretrained_model_config=resources_pb2.PretrainedModelConfig(),
         inference_compute_info=self.inference_compute_info,
-    )
+        output_info=resources_pb2.OutputInfo(params_specs=inference_param_proto))
```
ackizilkale, line 555: this one as well. not sure why it is here.

tests/runners/test_num_threads_config.py
```diff
@@ -0,0 +1,66 @@
...
+
+from clarifai.runners.models.model_builder import ModelBuilder
+
+
+@pytest.fixture
+def my_tmp_path(tmp_path):
+  return tmp_path
+
+
+@pytest.mark.parametrize("num_threads", [-1, 0, 3, 1.5, "a", None])
```
ackizilkale, line 15: nice, good test.

ackizilkale:
the original logic looks good. But there is some extra content I could not follow.

zeiler (approved):
nice!

PR #525, opened 2025-03-01
New model interface 4

deigen:
Changes the model interface to pythonic


clarifai/client/model_client.py
```diff
@@ -0,0 +1,422 @@
...
+        Initialize the model client.
+
+        Args:
+            stub: The gRPC stub for the model.
+            request_template: The template for the request to send to the model, including
+            common fields like model_id, model_version, cluster, etc.
+        '''
+    self.STUB = stub
+    self.request_template = request_template or service_pb2.PostModelOutputsRequest()
+    self._fetch_signatures()
```
zeiler, line 32: wonder if this should be lazy so that init of the ModelClient which I'm sure lots of people will have early in their code isn't having to make a predict call.

deigen: I was wondering the same thing, but it adds the functions to the model class instance, which is nice because you can run dir() or help() on the client in interactive sessions; they wouldn't have the functions installed otherwise.  I'll try making it load a little later, maybe on the first getattr miss and/or with an explicit call.

clarifai/runners/utils/method_signatures.py
```diff
@@ -0,0 +1,476 @@
...
+    tp = get_args(tp)[0]
+
+  return _normalize_data_type(tp), streaming
+
+
+def _normalize_data_type(tp):
+  # check if list, and if so, get inner type
+  if get_origin(tp) == list:
+    tp = get_args(tp)[0]
+    return List[_normalize_data_type(tp)]
```
luv-bansal, line 311: This should be updated to 
```
    if get_args(tp):
      return List[_normalize_data_type(get_args(tp)[0])]
```

currently it will fails for `  def predict(self, x: str, y:  Dict) -> List`

clarifai/runners/utils/method_signatures.py
```diff
@@ -0,0 +1,476 @@
...
+    pass  # not hashable type
+
+  raise TypeError(f'Unsupported type: {tp}')
+
+
+def _is_jsonable(tp):
+  if tp in (dict, list, tuple, str, int, float, bool, type(None)):
+    return True
+  if get_origin(tp) == list:
+    return _is_jsonable(get_args(tp)[0])
```
luv-bansal, line 352: same, it should be updated to

```
  if get_origin(tp) == list or get_origin(tp) == dict:
    return all(_is_jsonable(val) for val in get_args(tp))
  return False
```

luv-bansal:
@deigen  Some edge cases for which I was facing issues are:

Case 1:
```
class MyModel(ModelClass):

  @ModelClass.method
  def predict(self, x: str, y:  Dict) -> NamedFields:
    for key, value in y.items():
      x += key + str(value)

    return NamedFields(x=x, y=3)
```
Error
```
│ /home/luvbansal/anaconda3/envs/model-testing/lib/python3.10/site-packages/clarifai/runners/utils │
│ /method_signatures.py:39 in build_function_signature                                             │
│                                                                                                  │
│    36 │     build_variable_signature(p.name, p.annotation, p.default) for p in sig.parameters.   │
│    37   ]                                                                                        │
│    38   input_sigs, input_types, input_streaming = zip(*input_sigs)                              │
│ ❱  39   output_sig, output_type, output_streaming = build_variable_signature(                    │
│    40 │     'return', return_annotation, is_output=True)                                         │
│    41   # TODO: flatten out "return" layer if not needed                                         │
│    42                                                                                            │
│                                                                                                  │
│ /home/luvbansal/anaconda3/envs/model-testing/lib/python3.10/site-packages/clarifai/runners/utils │
│ /method_signatures.py:102 in build_variable_signature                                            │
│                                                                                                  │
│    99 │   raise ValueError(f'Invalid variable name: {name}')                                     │
│   100                                                                                            │
│   101   # get fields for each variable based on type                                             │
│ ❱ 102   tp, streaming = _normalize_type(annotation)                                              │
│   103                                                                                            │
│   104   #var = resources_pb2.VariableSignature()   # TODO                                        │
│   105   sig = _VariableSignature()  #for now                                                     │
│                                                                                                  │
│ /home/luvbansal/anaconda3/envs/model-testing/lib/python3.10/site-packages/clarifai/runners/utils │
│ /method_signatures.py:304 in _normalize_type                                                     │
│                                                                                                  │
│   301   if streaming:                                                                            │
│   302 │   tp = get_args(tp)[0]                                                                   │
│   303                                                                                            │
│ ❱ 304   return _normalize_data_type(tp), streaming                                               │
│   305                                                                                            │
│   306                                                                                            │
│   307 def _normalize_data_type(tp):                                                              │
│                                                                                                  │
│ /home/luvbansal/anaconda3/envs/model-testing/lib/python3.10/site-packages/clarifai/runners/utils │
│ /method_signatures.py:346 in _normalize_data_type                                                │
│                                                                                                  │
│   343   except TypeError:                                                                        │
│   344 │   pass  # not hashable type                                                              │
│   345                                                                                            │
│ ❱ 346   raise TypeError(f'Unsupported type: {tp}')                                               │
│   347                                                                                            │
│   348                                                                                            │
│   349 def _is_jsonable(tp):                                                                      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: Unsupported type: <class 'clarifai.runners.utils.data_types.NamedFields'>
```
Case 2:
```
class MyModel(ModelClass):

  @ModelClass.method
  def predict(self, x: str, y:  Dict[str, str]) -> Tuple:
    for key, value in y.items():
      x += key + value

    return x, x

```
error:
```
 /home/luvbansal/anaconda3/envs/model-testing/lib/python3.10/site-packages/clarifai/runners/utils │
│ /method_signatures.py:39 in build_function_signature                                             │
│                                                                                                  │
│    36 │     build_variable_signature(p.name, p.annotation, p.default) for p in sig.parameters.   │
│    37   ]                                                                                        │
│    38   input_sigs, input_types, input_streaming = zip(*input_sigs)                              │
│ ❱  39   output_sig, output_type, output_streaming = build_variable_signature(                    │
│    40 │     'return', return_annotation, is_output=True)                                         │
│    41   # TODO: flatten out "return" layer if not needed                                         │
│    42                                                                                            │
│                                                                                                  │
│ /home/luvbansal/anaconda3/envs/model-testing/lib/python3.10/site-packages/clarifai/runners/utils │
│ /method_signatures.py:108 in build_variable_signature                                            │
│                                                                                                  │
│   105   sig = _VariableSignature()  #for now                                                     │
│   106   sig.name = name                                                                          │
│   107                                                                                            │
│ ❱ 108   _fill_signature_type(sig, tp)                                                            │
│   109                                                                                            │
│   110   sig.streaming = streaming                                                                │
│   111                                                                                            │
│                                                                                                  │
│ /home/luvbansal/anaconda3/envs/model-testing/lib/python3.10/site-packages/clarifai/runners/utils │
│ /method_signatures.py:142 in _fill_signature_type                                                │
│                                                                                                  │
│   139 │   for inner_type in get_args(tp):                                                        │
│   140 │     #inner_sig = sig.type_args.add()                                                     │
│   141 │     sig.type_args.append(inner_sig := _VariableSignature())                              │
│ ❱ 142 │     _fill_signature_type(inner_sig, inner_type)                                          │
│   143 │   return                                                                                 │
│   144                                                                                            │
│   145   if get_origin(tp) == list:                                                               │
│                                                                                                  │
│ /home/luvbansal/anaconda3/envs/model-testing/lib/python3.10/site-packages/clarifai/runners/utils │
│ /method_signatures.py:153 in _fill_signature_type                                                │
│                                                                                                  │
│   150 │   _fill_signature_type(inner_sig, inner_type)                                            │
│   151 │   return                                                                                 │
│   152                                                                                            │
│ ❱ 153   raise TypeError(f'Unsupported type: {tp}')                                               │
│   154                                                                                            │
│   155                                                                                            │
│   156 def serializer_from_signature(signature):                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: Unsupported type: ()
```

clarifai/runners/models/model_class.py
```diff
@@ -1,41 +1,269 @@
...
 class ModelClass(ABC):
+  '''
+  Base class for model classes that can be run as a service.
+
+  Define predict, generate, or stream methods using the @ModelClass.method decorator.
+
+  Example:
+
+    from clarifai.runners.model_class import ModelClass, methods
+    from clarifai.runners.utils.data_types import Input, Stream
```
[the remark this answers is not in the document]
deigen: Thanks --- it is now NamedFields, updated.

deigen:
Thanks, I fixed all of these (btw, please feel free to add tests in the unittest file and fix as these come up, too).

For the last two cases:  NamedFields needs to have types, but I added a more helpful error message for this by checking for it in normalize_types; there is also a test case now for the error message.  Tuple probably should too, but it's reasonable to map it to json like list, since they're often both used for untyped returns.  On the client side we'll get a list back instead of a tuple, but that's probably ok.

deigen:
actually for case 2, for now I think we should require tuple to have types, and use List for untyped multiple values.  if that is too cumbersome though, add a case that checks for untyped tuples and handles similarly to untyped lists.

clarifai/cli/model.py
```diff
@@ -76,6 +76,31 @@ def download_checkpoints(model_path, out_path, stage):
     required=False,
     default=".",
 )
+@click.option(
```
zeiler, line 79: if you're creating the sigs upfront I don't think we need the CLI anymore right?

clarifai/client/model.py
```diff
@@ -47,6 +49,9 @@ def __init__(self,
                pat: str = None,
                token: str = None,
                root_certificates_path: str = None,
+               compute_cluster_id: str = None,
+               nodepool_id: str = None,
+               deployment_id: str = None,
```
zeiler, line 54: we still need a way to specify the user_id of the model separate from the user_id of the deployment. I don't know why user_id goes into kwargs either, it makes looking at the code very confusion what's supported.

clarifai/runners/dockerfile_template/Dockerfile.template
```diff
@@ -44,6 +44,9 @@ ENV PYTHONPATH=${PYTHONPATH}:/home/nonroot/main \
     CLARIFAI_COMPUTE_CLUSTER_ID=${CLARIFAI_COMPUTE_CLUSTER_ID} \
     CLARIFAI_API_BASE=${CLARIFAI_API_BASE:-https://api.clarifai.com}
 
+# # Write out the model function signatures
```
zeiler, line 47: i think we can remove this

clarifai/runners/models/model_builder.py
```diff
@@ -531,10 +598,12 @@ def hf_labels_to_config(self, labels, config_file):
     logger.info(f"Updated config.yaml with {len(concepts)} concepts.")
 
   def get_model_version_proto(self):
-
+    signatures = self.get_method_signatures()
+    # TODO: update this to `method_signatures` field when it's available in the API
     model_version_proto = resources_pb2.ModelVersion(
         pretrained_model_config=resources_pb2.PretrainedModelConfig(),
         inference_compute_info=self.inference_compute_info,
+        method_signature=signatures,
```
zeiler, line 606: nice, this is amazing, @patricklundquist @sainivedh all we need to do now is add this method signatures to the API to record it in the DB. Then that will be returned by the API for the models to use in the UI and clients in a GetModelOutputInfo type of call I think it may be called or GetModelVersion perhaps. @sainivedh maybe you can hammer that out today and then lundquist will merge in and we can merge in this PR too to have the new method signatures to play with!

clarifai/runners/models/model_class.py
```diff
@@ -1,41 +1,292 @@
...
+  def _handle_get_signatures_request(self) -> service_pb2.MultiOutputResponse:
+    methods = self._get_method_info()
+    signatures = {method.name: method.signature for method in methods.values()}
+    resp = service_pb2.MultiOutputResponse(status=status_pb2.Status(code=status_code_pb2.SUCCESS))
+    output = resp.outputs.add()
+    output.status.code = status_code_pb2.SUCCESS
+    output.data.text.raw = signatures_to_json(signatures)
+    return resp
+
+  def _batch_predict(self, method, inputs: List[Dict[str, Any]]) -> List[Any]:
```
zeiler, line 68: i don't understand how to use these batch methods from the client, is there an example of that?

clarifai/runners/utils/data_types.py
```diff
@@ -0,0 +1,427 @@
...
+
+  def to_proto(self) -> RegionProto:
+    return self.proto
+
+  @classmethod
+  def from_proto(cls, proto: RegionProto) -> "Region":
+    return cls(proto)
+
+
+class Image(MessageData):
```
zeiler, line 210: in other parts of our SDK outside the runners package don't we already have some of these core types defined or is it just Input?

clarifai/runners/utils/method_signatures.py
```diff
@@ -0,0 +1,477 @@
...
+  # method_signature.method_type = method_type
+  method_signature.description = inspect.cleandoc(func.__doc__ or '')
+  # method_signature.annotations_json = json.dumps(_get_annotations_source(func))
+
+  method_signature.input_fields.extend(input_sigs)
+  method_signature.output_fields.append(output_sig)
+  return method_signature
+
+
+# def _get_annotations_source(func):
```
zeiler, line 72: delete?

tests/runners/test_runners.py
```diff
@@ -3,6 +3,7 @@
 #
```
zeiler, line 3: not sure if this file is relevant but are there any tests where the model is the old proto model but the client is the new SDK interface calling predict()? that would help us know if migrating old triton models to the proto interface and not this new interface would be fine.

clarifai/runners/models/model_builder.py
sainivedh: Started -> https://clarifai.atlassian.net/browse/EAGLE-5879

clarifai/runners/dockerfile_template/Dockerfile.template
luv-bansal: Removed

clarifai/runners/models/model_class.py
luv-bansal: Refer this section in the Readme: 

https://github.com/Clarifai/examples/blob/76d0c27a4deebc3c5150482b45945b64b6d81915/models/model_upload/README.md#dynamic-batch-prediction-handling

clarifai/cli/model.py
luv-bansal: yeah we don't need this CLI now, sometime for testing I use this CLI command

clarifai/client/model.py
luv-bansal: Yeah, currently user need to have `user_id ` in env variable with `CLARIFAI_USER_ID` to use the deployment

luv-bansal: One way is to use `model_user_id` for model's user_id, while `user_id` as an deployment user_id, WDT?

clarifai/runners/utils/data_types.py
luv-bansal: Yeah I think we have test for this

tests/runners/test_runners.py
luv-bansal: Yeah already have test for this

clarifai/runners/utils/data_types.py
sainivedh: I think it's just `Input` class

clarifai/runners/models/model_builder.py
```diff
@@ -82,7 +97,30 @@ def create_model_instance(self, load_model=True):
...
+    original_import = builtins.__import__
+
+    def custom_import(name, globals=None, locals=None, fromlist=(), level=0):
+
+      # Allow standard libraries and clarifai
+      if self._is_standard_or_clarifai(name):
+        return original_import(name, globals, locals, fromlist, level)
+
+      # Mock all third-party imports to avoid ImportErrors or other issues
+      return MagicMock()
```
abhash-ai, line 110: Should there be a warning here too? In case mocking could potentially result in errors?

ackizilkale (approved):
lgtm

PR #526, opened 2025-03-03
[PR-395] fix docker image name + skip dockerfile when test

phatvo9:
### What

Test-locally with `--mode container` failed if any char of `model-id` is in upper case.
Can't test -locally with custom dockerfile as it auto generates new dockerfile everytime run cli

### Why

*

### How

* Lower case model-id when building image
* add `skip_dockerfile` flag

### Tests

*

### Notes

*


mogith-pn (approved):
Looks good to me. LGTM

PR #529, opened 2025-03-04
[PR-394] Specify model_path by positional arg

phatvo9:
### What 

Specify model_path by positional arg

### Why

* make our cli shorter

### How

* `model_path` as positional arg with default value is current directory
```
# old command
clarifai model upload --model_path ./your/model/path

# new command
clarifai model upload ./your/model/path

# or
cd ./your/mode/path
clarifai model upload

```

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/cli/model.py
+++ b/clarifai/cli/model.py
@@ -9,11 +9,7 @@ def model():
 
 
 @model.command()
-@click.option(
-    '--model_path',
-    type=click.Path(exists=True),
-    required=True,
-    help='Path to the model directory.')
+@click.argument("model_path", type=click.Path(exists=True), required=False, default=".")
 @click.option(
     '--stage',
     required=False,
@@ -36,11 +32,12 @@ def upload(model_path, stage, skip_dockerfile):
 
 
 @model.command()
-@click.option(
-    '--model_path',
+@click.argument(
+    "model_path",
     type=click.Path(exists=True),
-    required=True,
-    help='Path to the model directory.')
+    required=False,
+    default=".",
+)
 @click.option(
     '--out_path',
     type=click.Path(exists=False),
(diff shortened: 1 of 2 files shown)


mogith-pn (commented):
Looks good for me. Just one suggestion.
@phatvo9 Can we have docstring description for the `model_path` , as since click.argument doesn't support the `help` parameter.

clarifai/cli/model.py
```diff
@@ -9,11 +9,7 @@ def model():
 
 
 @model.command()
-@click.option(
-    '--model_path',
-    type=click.Path(exists=True),
-    required=True,
-    help='Path to the model directory.')
+@click.argument("model_path", type=click.Path(exists=True), required=False, default=".")
```
zeiler, line 12: i see mogith's suggestion to add help but don't see it in these?

phatvo9: click.argument does not have `help`. Need to define in docstring. Example of `clarifai model upload --help`
```
Usage: clarifai model upload [OPTIONS] [MODEL_PATH]

  Upload a model to Clarifai.

  MODEL_PATH: Path to the model directory. If not specified, the current
  directory is used by default.

Options:
  --stage [runtime|build|upload]  The stage we are calling download
                                  checkpoints from. Typically this would
                                  "upload" and will download checkpoints if
                                  config.yaml checkpoints section has when set
                                  to "upload". Other options include "runtime"
                                  to be used in load_model or "upload" to be
                                  used during model upload. Set this stage to
                                  whatever you have in config.yaml to force
                                  downloading now.  [default: upload]
  --skip_dockerfile               Flag to skip generating a dockerfile so that
                                  you can manually edit an already created
                                  dockerfile.
  --help                          Show this message and exit.


```

zeiler (approved):

clarifai/cli/model.py
zeiler: ah ok, that works, thanks!

PR #540, opened 2025-03-25
Added Validation for CLI config

srikanthbachala20:
### Why

* Run validation for CLI config only when necessary

### How

* Added a utility function to validate if config data exist
* Called this utility function in commands where config is necessary

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/cli/base.py
+++ b/clarifai/cli/base.py
@@ -43,17 +43,10 @@ def cli(ctx):
   """Clarifai CLI"""
   ctx.ensure_object(dict)
   config_path = f"{os.getenv('HOME')}/.clarifai/config.yaml"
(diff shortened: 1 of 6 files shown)


ackizilkale:
@srikanthbachala20 what will happen on the pod do you think?

ackizilkale:
Cause I am not sure we have write access on the pod there.

srikanthbachala20:
> @srikanthbachala20 what will happen on the pod do you think?

Previously if the config data is missing then it will automatically redirect to login command. But I have removed that part and just added a error message to login if the config data is missing for the subcommands which require the config data. For subcommands like model upload, download checkpoints don't need to config. So, they won't throw error now.

ackizilkale (approved):

PR #542, opened 2025-03-26
[PR-470]: Added Regex Patterns to Filter Checkpoint Files to Download

srikanthbachala20:
### Why

* To download only required files from huggingface repo

### How

* By adding list of regex patterns to allow & ignore to model config

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/runners/utils/loader.py
+++ b/clarifai/runners/utils/loader.py
@@ -6,6 +6,7 @@ import shutil
 
 import requests
 
+from clarifai.runners.utils.const import CONCEPTS_REQUIRED_MODEL_TYPE
 from clarifai.utils.logging import logger
 
 
@@ -13,9 +14,10 @@ class HuggingFaceLoader:
 
   HF_DOWNLOAD_TEXT = "The 'huggingface_hub' package is not installed. Please install it using 'pip install huggingface_hub'."
 
-  def __init__(self, repo_id=None, token=None):
+  def __init__(self, repo_id=None, token=None, model_type_id=None):
     self.repo_id = repo_id
     self.token = token
+    self.clarifai_model_type_id = model_type_id
     if token:
       if self.validate_hftoken(token):
         try:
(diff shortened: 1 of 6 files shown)


mogith-pn (commented):
Looks good, I'm going to test this branch for the model now and will approve.

clarifai/runners/utils/loader.py
```diff
@@ -121,8 +130,17 @@ def validate_download(self, checkpoint_path: str):
     repo_files = list_repo_files(self.repo_id, token=self.token)
 
     self.ignore_patterns = self._get_ignore_patterns()
-    # Get the list of files on the repo that are not ignored
+    if ignore_file_patterns:
+      self.ignore_patterns.extend(ignore_file_patterns)
```
mogith-pn, line 134: This condition is failing for the repos which doesn't have the models as safetensors format. 
For example - TheBloke/Llama-2-7B-GGUF
Inside `the _get_ignore_patterns` function -
```
def _get_ignore_patterns(self):
    # check if model exists on HF
    try:
      from huggingface_hub import list_repo_files
    except ImportError:
      raise ImportError(self.HF_DOWNLOAD_TEXT)
    # Get the list of files on the repo that are not ignored
    repo_files = list_repo_files(self.repo_id, token=self.token)
    self.ignore_patterns = None
    if any(f.endswith(".safetensors") for f in repo_files):
      self.ignore_patterns = [
          "**/original/*", "**/*.pth", "**/*.bin", "*.pth", "*.bin", "**/.cache/*"
      ]
    return self.ignore_patterns
```
self.ignore_patterns were set to `None` so eventually this throws an error.

mogith-pn, line 134:
```suggestion
if self.ignore_patterns is not None:
     self.ignore_patterns.extend(ignore_file_patterns)
else:
    self.ignore_patterns=ignore_file_patterns
```

mogith-pn (changes requested):
Somehow when I still testing model download locally, I'm getting errors after the model is downloaded. I'm guessing it is failing while validating the no of files after downloaded. Can you check it for multiple repos locally and let me know. 
The repos that I tested with - 
```
TheBloke/Llama-2-7B-GGUF
meta-llama/Llama-3.2-1B-Instruct
```

clarifai/runners/utils/loader.py
srikanthbachala20: Thanks Mogith

srikanthbachala20: Done

mogith-pn (approved):
Thanks for iterating. I have tested locally as well. Looks good. 
LGTM.

PR #543, opened 2025-03-26
Removed HF loader config.json validation for all clarifai Model type ids

mogith-pn:
### Why

* In SDK while validating HF checkpoints from config.yaml, we are doing a check to validate whether `config.json` present inside the `repo_id`.

* This is to essentially fetch concepts for image classification and segmentation type tasks, which is irrelevant for LLM and VLM models. Due to this some GGUF models from HF couldn’t able to be uploaded since it doesn’t have config.json in the repo.

### How

* By just using the validation step to check if the current model type id is in `CONCEPTS_REQUIRED_MODEL_TYPE` list and only checks for `config.json` if it passes the condition.

### Tests

* Not added at this point. But will add in future with dummy model with model type id in `CONCEPTS_REQUIRED_MODEL_TYPE`  without` config.json` and a lightweight `text-to-text` model without `config.json`.

### DIFF
--- a/clarifai/runners/models/model_builder.py
+++ b/clarifai/runners/models/model_builder.py
@@ -480,6 +480,7 @@ class ModelBuilder:
     if not self.config.get("checkpoints"):
       logger.info("No checkpoints specified in the config file")
       return path
+    clarifai_model_type_id = self.config.get('model').get('model_type_id')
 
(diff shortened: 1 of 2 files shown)


srikanthbachala20 (approved):
LGTM

PR #544, opened 2025-03-27
multiple requests according to model type

mansi-k:
Ticket1: https://clarifai.atlassian.net/browse/PR-468

Ticket2: https://clarifai.atlassian.net/browse/PR-471

For multimodal, the image and the text in the test is unrelated. Made them related to make sure the image is taken into consideration by the model 

Only image url seems to be considered. Enabled passing imagebytes too, as some models do image processing before prediction

Also added test for multimodal models with only text in the input

Logged the last token in the streamed response


clarifai/runners/models/model_run_locally.py
```diff
@@ -19,503 +22,702 @@
...
+
+        model_version_proto = self.builder.get_model_version_proto()
+        model_version_proto.id = "model_version"
+        image_url = "https://samples.clarifai.com/metro-north.jpg"
+        image_path = "../../static/metro-north.jpg"
+
+        default_request = service_pb2.PostModelOutputsRequest(
+            model=resources_pb2.Model(model_version=model_version_proto),
+            inputs=[
+                resources_pb2.Input(
```
abhash-ai, line 106: Dont need a default request, can just use `_build_text_to_text_requests` as default request no?

clarifai/runners/models/model_run_locally.py
```diff
@@ -19,503 +22,702 @@
...
+        )
+
+        def image_to_base64(image_path):
+            with open(image_path, "rb") as image_file:
+                base64_img = base64.b64encode(image_file.read())
+            return base64_img
+
+        def _build_text_to_text_requests():
+            requests = [
+                service_pb2.PostModelOutputsRequest(
```
abhash-ai, line 132: **Nit:** I think we can reduce duplication by creating a function: CreateRequest(model_version_proto, image_input, text_input), where you can pass in the Image, Text and ModelVersionProto as params?

Would make it reusable elsewhere too

clarifai/runners/models/model_run_locally.py
```diff
@@ -19,503 +22,702 @@
...
+
+        try:
+            generate_response = model.generate(request)
+        except NotImplementedError:
+            logger.info("Model does not implement generate() method.")
+        except Exception as e:
+            logger.error(f"Model Generation failed: {e}")
+            traceback.print_exc()
+            generate_response = service_pb2.MultiOutputResponse(
+                status=status_pb2.Status(
```
abhash-ai, line 258: Unrelated to this PR, but shouldn't the predict / generate stream functions themselves perform this fallback to calling `MultiOutputResponse` API?

clarifai/runners/models/model_run_locally.py
```diff
@@ -19,503 +22,702 @@
...
+                    # Kill the process if it doesn't terminate after 5 seconds
+                    process.kill()
+
+    # run the model server
+    def run_model_server(self, port=8080):
+        """Run the Clarifai Runners's model server."""
+
+        command = [
+            self.python_executable,
+            "-m",
```
abhash-ai, line 367: Previously was better?

luv-bansal (commented):
Test-locally will work differently in the pythonic interface, so not sure if we need this

luv-bansal:
@ackizilkale I'm sure we don't need this anymore, so closing this PR

PR #554, opened 2025-04-08
clean up some logging

zeiler:
fix some imports around logging


sainivedh (approved):
Perhaps we should document the reason why two different loggers are currently needed and create any necessary cleanup tickets for this?

PR #558, opened 2025-04-17
Update region

phatvo9:
### What
1. Adding these to `Region`:
* mask
* point
* track_id
2. Update test-locally with removing --model_path (already available on master)


ackizilkale (approved):
looks good to me.

clarifai/runners/utils/data_types/data_types.py
```diff
@@ -176,43 +183,191 @@ def from_proto(cls, proto: ConceptProto) -> "Concept":
...
 
   @property
   def box(self) -> List[float]:
     bbox = self.proto.region_info.bounding_box
-    return [bbox.left_col, bbox.top_row, bbox.right_col, bbox.bottom_row]  # x1, y1, x2, y2
+    # x1, y1, x2, y2
+    return [bbox.left_col, bbox.top_row, bbox.right_col, bbox.bottom_row]
 
   @box.setter
   def box(self, value: List[float]):
```
mogith-pn, line 318:
```suggestion
  def box(self, value: List[Union[int, float]]):
```

mogith-pn: Just this suggestion. Everything else looks good to me. LGTM.

phtvo: It has to be float cause attributes of bbox proto are float.

ackizilkale (approved):

clarifai/runners/utils/data_types/__init__.py
luv-bansal, line 1: don't we need to import `Point` and `Mask` too?

tests/runners/test_model_signatures.py
luv-bansal, line 1: @phatvo9 can you add test cases for `Point` and `Mask` too here? like we have for other datatypes

luv-bansal (commented):
lgtm, just we need test cases for Mark and Point

clarifai/runners/utils/data_types/__init__.py
phatvo9: Just rework on these. I don't think we need a Mask or Point outside Region. Move them to Region and removed their classes.

tests/runners/test_model_signatures.py
phatvo9: sure. added.

luv-bansal (approved):
Looks good!!!

PR #560, opened 2025-04-22
remove rich from req

sainivedh:
### Why

* Remove `rich` as a necessary package in `requirements.txt`

### How

* Prompt user to install `rich` when required, such as `workflow graph display` and `dataset upload status`

### Tests

* Manual test by observing model upload logs

### Notes

*


sanjaychelliah (approved):
LGTM

PR #562, opened 2025-04-23
PR-500 simplified client wrapper functions

mogith-pn:
### Why

* https://clarifai.atlassian.net/browse/PR-500
* The client wrapper function in model upload code, still needed some abstraction for converting images/video/audio into openai compatible format before sending to chat function.

### How

* Since it can be abstacted into the runner data utils function, moved those preprocessing steps into SDK.

### Tests

* PR tested with model upload into production after making changes - https://clarifai.com/mogith-p-n/test-co/models/smollm
* Uploaded model code  - https://github.com/Clarifai/model-uploads/tree/smollm-model/SmolLM/SmolLM2-1.7B-Instruct
* Tested all the functions after deployment.

### DIFF
--- a/clarifai/runners/utils/data_utils.py
+++ b/clarifai/runners/utils/data_utils.py
@@ -1,24 +1,24 @@
 from io import BytesIO
-from typing import List
+from typing import Dict, List
 
 from clarifai_grpc.grpc.api import resources_pb2
 from clarifai_grpc.grpc.api.resources_pb2 import ModelTypeEnumOption
 from clarifai_grpc.grpc.api.resources_pb2 import ModelTypeField as InputFieldProto
 from clarifai_grpc.grpc.api.resources_pb2 import ModelTypeRangeInfo
-from PIL import Image
+from PIL import PILImage
 
-from clarifai.runners.utils.data_types import MessageData
+from clarifai.runners.utils.data_types import Audio, Image, MessageData, Video
 
 
-def image_to_bytes(img: Image.Image, format="JPEG") -> bytes:
+def image_to_bytes(img: PILImage.Image, format="JPEG") -> bytes:
   buffered = BytesIO()
   img.save(buffered, format=format)
   img_str = buffered.getvalue()
   return img_str
 
 
-def bytes_to_image(bytes_img) -> Image.Image:
-  img = Image.open(BytesIO(bytes_img))
+def bytes_to_image(bytes_img) -> PILImage.Image:
+  img = PILImage.open(BytesIO(bytes_img))
   return img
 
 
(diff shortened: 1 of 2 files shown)


sanjaychelliah (approved):

clarifai/runners/utils/data_utils.py
```diff
@@ -58,6 +58,126 @@ def is_openai_chat_format(messages):
   return True
 
 
+def build_openai_chat_format(prompt: str, image: Image, images: List[Image], audio: Audio,
```
luv-bansal, line 61: Not sure, do you think we could move this to [here](https://github.com/Clarifai/clarifai-python/blob/master/clarifai/runners/utils/openai_convertor.py)

luv-bansal (commented):
Looks good, but not sure , should we add `build_openai_chat_format` function to `utils/openai_convertor.py` file?

clarifai/runners/utils/data_utils.py
luv-bansal, line 61: also should be rename this function it to `build_openai_messages`?

clarifai/runners/utils/data_types/data_types.py
```diff
@@ -395,6 +395,14 @@ def to_pil(self) -> PILImage.Image:
...
         return PILImage.open(io.BytesIO(self.proto.base64))
 
+    def to_base64_str(self) -> str:
+        if isinstance(self.proto.base64, bytes):
+            return self.proto.base64.decode('utf-8')
+        elif isinstance(self.proto.base64, str):
+            return self.proto.base64
+        elif not self.proto.base64:
+            raise ValueError("Image has no bytes")
+
```
luv-bansal, line 405: don't we need to do `base64.b64encode(self.proto.base64).decode('utf-8')` for converting bytes to `base64_str`? because `self.proto.base64` is `bytes` not `base64`

clarifai/runners/utils/data_types/data_types.py
```diff
@@ -466,6 +474,14 @@ def __repr__(self) -> str:
...
         return self.proto
 
+    def to_base64_str(self) -> str:
+        if isinstance(self.proto.base64, bytes):
+            return self.proto.base64.decode('utf-8')
+        elif isinstance(self.proto.base64, str):
+            return self.proto.base64
+        elif not self.proto.base64:
+            raise ValueError("Audio has no bytes")
+
```
luv-bansal, line 484: similarly here for all `to_base64_str` methods

luv-bansal (commented):
I feel like `to_base64_str` implementation is currently wrong, because I'm sure `self.proto.base64 ` is `bytes` and not a base64-encoded bytes. I know this confusion in naming is for long time in Clarifai

clarifai/runners/utils/openai_convertor.py
```diff
@@ -164,3 +168,102 @@ def openai_to_hf_messages(openai_messages):
...
+          videos (List[Video]): List of Clarifai Video objects.
+          messages (List[Dict]): List of chat messages.
+      Returns:
+          List[Dict]: Formatted chat messages.
+    """
+
+    openai_messages = []
+    # Add previous conversation history
+    if messages:
+        openai_messages.extend(messages)
```
luv-bansal, line 201: I think it's good to add `is_openai_chat_format(messages)` flag check for `messages` like `if  messages and is_openai_chat_format(messages)`

clarifai/runners/utils/data_types/data_types.py
```diff
@@ -475,12 +481,18 @@ def to_proto(self) -> AudioProto:
...
-            return self.proto.base64
-        elif not self.proto.base64:
+        if not self.proto.base64:
             raise ValueError("Audio has no bytes")
+        if isinstance(self.proto.base64, str):
+            return self.proto.base64
+        if isinstance(self.proto.base64, bytes):
+            try:
+                # trying direct decode (if already a base64 bytes)
+                return self.proto.base64.decode('utf-8')
```
luv-bansal, line 491: I think this always works i.e it never goes to except block because `self.proto.base64` is bytes and and `self.proto.base64.decode('utf-8')` always works. 

I just asked gpt, how to differential `bytes` and `base64` in python
```python
def is_base64(s):
    try:
        if isinstance(s, str):
            s_bytes = s.encode('utf-8')
        elif isinstance(s, bytes):
            s_bytes = s
        else:
            return False
        return base64.b64encode(base64.b64decode(s_bytes)) == s_bytes
    except Exception:
        return False
```

luv-bansal (approved):

clarifai/runners/utils/data_types/data_types.py
mogith-pn: As discussed, this condition will be applicable when we do request.get("public url") and it returns a raw binary data. So added an raise exception block to fallback when all conditions fail.

PR #564, opened 2025-04-24
Removed duplicate model downloads and improved error logging for gated HF repo.

mogith-pn:
### Why

* Model download is happening twice while performing `test-locally` through CLI.
* Error logging is not clearly showing when the model upload is failing due to gated repository access issue. 

### How

* Adding checks to ensure repo is valid and user has access to it before the build process to reduce overhead.
* Removed total block which is doing redundant download locally.

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/runners/models/model_builder.py
+++ b/clarifai/runners/models/model_builder.py
@@ -287,10 +287,20 @@ class ModelBuilder:
       if loader_type == "huggingface" and hf_token:
         is_valid_token = HuggingFaceLoader.validate_hftoken(hf_token)
         if not is_valid_token:
+          logger.info(
+              "Continuing without Hugging Face token for validating config in model builder.")
+        if is_valid_token:
+          is_repo_access_valid = HuggingFaceLoader.validate_hf_repo_access(
+              repo_id=(self.config.get("checkpoints").get("repo_id")), token=hf_token)
+        else:
+          is_repo_access_valid = HuggingFaceLoader.validate_hf_repo_access(
+              self.config.get("checkpoints").get("repo_id"))
+
+        if not is_repo_access_valid:
           logger.error(
-              "Invalid Hugging Face token provided in the config file, this might cause issues with downloading the restricted model checkpoints."
+              f"Invalid Hugging Face repo access for repo {self.config.get('checkpoints').get('repo_id')}. Please check your repo and try again."
           )
-          logger.info("Continuing without Hugging Face token")
+          sys.exit("Invalid Hugging Face repo access, exiting.")
 
     num_threads = self.config.get("num_threads")
     if num_threads or num_threads == 0:
--- a/clarifai/runners/models/model_run_locally.py
+++ b/clarifai/runners/models/model_run_locally.py
@@ -404,9 +404,6 @@ def main(model_path,
   manager = ModelRunLocally(model_path)
   # get whatever stage is in config.yaml to force download now
   # also always write to where upload/build wants to, not the /tmp folder that runtime stage uses
-  _, _, _, when, _, _ = manager.builder._validate_config_checkpoints()
-  manager.builder.download_checkpoints(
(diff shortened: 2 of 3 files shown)


clarifai/runners/models/model_builder.py
```diff
@@ -287,10 +287,20 @@ def _validate_config(self):
...
         is_valid_token = HuggingFaceLoader.validate_hftoken(hf_token)
         if not is_valid_token:
+          logger.info(
+              "Continuing without Hugging Face token for validating config in model builder.")
+        if is_valid_token:
+          is_repo_access_valid = HuggingFaceLoader.validate_hf_repo_access(
+              repo_id=(self.config.get("checkpoints").get("repo_id")), token=hf_token)
+        else:
+          is_repo_access_valid = HuggingFaceLoader.validate_hf_repo_access(
+              self.config.get("checkpoints").get("repo_id"))
```
abhash-ai, line 297: **nit:** could be done in a single if-else:
if is_valid_token:
   ....
else: 
   ....

clarifai/runners/models/model_builder.py
```diff
@@ -287,10 +287,20 @@ def _validate_config(self):
       if loader_type == "huggingface" and hf_token:
         is_valid_token = HuggingFaceLoader.validate_hftoken(hf_token)
         if not is_valid_token:
+          logger.info(
+              "Continuing without Hugging Face token for validating config in model builder.")
```
abhash-ai, line 291: could be an error or warning?

clarifai/runners/models/model_builder.py
```diff
@@ -287,10 +287,20 @@ def _validate_config(self):
...
+          is_repo_access_valid = HuggingFaceLoader.validate_hf_repo_access(
+              self.config.get("checkpoints").get("repo_id"))
+
+        if not is_repo_access_valid:
           logger.error(
-              "Invalid Hugging Face token provided in the config file, this might cause issues with downloading the restricted model checkpoints."
+              f"Invalid Hugging Face repo access for repo {self.config.get('checkpoints').get('repo_id')}. Please check your repo and try again."
           )
-          logger.info("Continuing without Hugging Face token")
+          sys.exit("Invalid Hugging Face repo access, exiting.")
```
abhash-ai, line 303: **nit:** Could say: does not have access to repo ...

clarifai/runners/models/model_builder.py
```diff
@@ -287,10 +287,20 @@ def _validate_config(self):
       if loader_type == "huggingface" and hf_token:
         is_valid_token = HuggingFaceLoader.validate_hftoken(hf_token)
         if not is_valid_token:
+          logger.info(
```
abhash-ai, line 290: **nit:** alternative implementation: 
Move the `HuggingFaceLoader.validate_hftoken(hf_token)` check to `HuggingFaceLoader.validate_hf_repo_access`. 

This way here you just have a single call to `HuggingFaceLoader.validate_hf_repo_access`, and no checks to validate_hf_token

abhash-ai (changes requested):
minor changes

clarifai/runners/models/model_builder.py
mogith-pn: This is not necessarily a warning/error, Model download is possible without HF_TOKEN.

clarifai/runners/models/model_builder.py
mogith-pn: Both are classmethods. 
There's a reason why both has to be done separate. Basically the first check is to ensure the user is not giving random token value or empty string while uploading the model, This happens when you copy examples and upload models. 
Second is to let the user know whether the download failed due to the repo is gated. Means the token must be valid but he has to sign an acknowledgement in the model page to access it further. So this ensures right feedback is returned to the user.

luv-bansal (approved):
looks good

PR #566, opened 2025-04-25
[PR-556]: Adding usage of subclasses of ModelClass in Model Upload

srikanthbachala20:
### Why

* To use special subclasses like VisualDetectorClass or VisualClassifierClass in Model Upload

### How

* Use indirectly inherited classes (like VisualDetectorClass or VisualClassifierClass) in model upload

### Tests

*

### Notes

*


luv-bansal (approved):
Looks good

PR #576, opened 2025-05-07
[EAGLE-5987]: fix another place where we need to use integers

zeiler:


luv-bansal:
@zeiler This looks good, so merging this PR.

PR #579, opened 2025-05-08
fix clarifai model predict

rizzip:
### Why

* `clarifai model predict` fails due to additional args passed to `predict_by*`

### How

* Pass the compute args to the Model instead

### Tests

* The args error no longer appears and the call either waits for deployment or fails in other valid ways

### Notes

*


luv-bansal (commented):
@srikanthbachala20 already fixed `clarifai model predict` CLI, I don't think we need this anymore. @rizzip I'm closing this PR

PR #581, opened 2025-05-09
[PR-544]AMD changes

luv-bansal:
### Why

*

### How

*

### Tests

*

### Notes

*


clarifai/runners/utils/const.py
```diff
@@ -4,14 +4,28 @@
 
 GIT_SHA = "b8ae56bf3b7c95e686ca002b07ca83d259c716eb"
 
+AMD_GIT_SHA = "b8ae56bf3b7c95e686ca002b07ca83d259c716eb"
```
patricklundquist, line 7: update to your new amd image git hash? 42938da8e33b0f37ee7db16b83631da94c2348b9

ackizilkale:
Isn't it better to write the logic in anticipation of TPU? Currently the logic is on `is_amd` boolean. We will have to change it when we want to do tpu

ackizilkale (approved):
See Patrick's comment above. Beyond that good to go

clarifai/runners/utils/const.py
luv-bansal: Done

PR #601, opened 2025-05-27
Improve openai class

luv-bansal:
### Why

*

### How

*

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/runners/models/openai_class.py
+++ b/clarifai/runners/models/openai_class.py
@@ -1,7 +1,7 @@
 """Base class for creating OpenAI-compatible API server."""
 
 import json
-from typing import Any, Dict, Iterator, List, Optional
+from typing import Any, Dict, Iterator
 
 from clarifai.runners.models.model_class import ModelClass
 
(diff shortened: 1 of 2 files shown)


clarifai/runners/models/openai_class.py
```diff
@@ -1,143 +1,222 @@
...
+            for chunk in self._process_streaming_request(**params):
+                if chunk.get('usage'):
+                    if chunk['usage'].get('prompt_tokens') and chunk['usage'].get(
+                        'completion_tokens'
+                    ):
+                        self.set_output_context(
+                            prompt_tokens=chunk['usage']['prompt_tokens'],
+                            completion_tokens=chunk['usage']['completion_tokens'],
+                        )
+                if (chunk.get('choices') and len(chunk['choices']) > 0) or (
```
zeiler, line 190: shouldn't we always just do a json.dumps(chunk) here and not have these iffs? Just let it always return what it has so we're always comopatible.

zeiler (approved):

PR #618, opened 2025-06-03
[PR-626] and update openai params

phatvo9:
* PR-626: default value of string param stored by json.dumps causing wrong display on UI (e.g. `"a"` -> `"\"a\""`)
* Allow all openai params go through in `openai_transport*` methods

### Why

*

### How

*

### Tests

*

### Notes

*

### DIFF
--- a/clarifai/runners/models/openai_class.py
+++ b/clarifai/runners/models/openai_class.py
@@ -35,97 +35,34 @@ class OpenAIModelClass(ModelClass):
                     "Subclasses must set the 'model' class attribute or ensure the client can list models"
                 ) from e
 
-    def _extract_request_params(self, request_data: Dict[str, Any]) -> Dict[str, Any]:
-        """Extract and validate common openai arguments parameters from the request data.
-
-        Args:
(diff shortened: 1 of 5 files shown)


clarifai/runners/models/openai_class.py
```diff
@@ -35,97 +35,34 @@ def __init__(self) -> None:
...
-
-        for param in optional_params:
-            if params.get(param) is not None:
-                completion_args[param] = params[param]
+            # Force to use usage
+            stream_options = params.pop("stream_options", {})
+            stream_options.update({"include_usage": True})
+            completion_args["stream_options"] = stream_options
+        completion_args["stream"] = stream
+        print(f"{completion_args=}")
```
luv-bansal, line 56: do we need to add `print` here?

luv-bansal (approved):
Looking good!! Nice, you also refracted the code.

clarifai/runners/models/openai_class.py
phatvo9: removed

PR #626, opened 2025-06-12
[EAGLE-6013]: use uv in the build process

zeiler:
Bumped to latest base images that have uv in them.


clarifai/runners/dockerfile_template/Dockerfile.template
```diff
@@ -46,6 +50,6 @@ ENV PYTHONPATH=${PYTHONPATH}:/home/nonroot/main \
 
 # Finally run the clarifai entrypoint to start the runner loop and local dev server.
 # Note(zeiler): we may want to make this a clarifai CLI call.
-ENTRYPOINT ["python", "-m", "clarifai.runners.server"]
+ENTRYPOINT ["uv", "python", "-m", "clarifai.runners.server"]
```
luv-bansal, line 53: I think `uv` not meant to be used as a prefix to python.

zeiler: true

luv-bansal (approved):

PR #637, opened 2025-06-18
PR-652 - App action for model upload

phatvo9:
### What

Show a terminal prompt asking users if they want to create a new app when the specified app does not exist.

### Why

*

### How

*

### Tests

*

### Notes

*


ackizilkale (approved):
Looks good to me. See if you agree with copilot comments :)

PR #672, opened 2025-07-03
Fix path windows

phatvo9:
### What
Failed to run clarifai cli on Windows because variable `HOME` does not contain windows home path.
Use `pathlib.Path.home()` instead.


sanjaychelliah (approved):
👍

PR #718, opened 2025-07-24
[PR-734] Use Method signature for local-runner

luv-bansal:
## Pull Request Overview

This PR introduces a new `patch_version` method to the Model class and integrates method signatures into the local runner workflow. The changes enable updating existing model versions with method signatures and configuration updates rather than always creating new versions.

- Adds a `patch_version` method to the Model class for updating existing model versions
- Updates the local runner to include method signatures when creating or updating model versions
- Refactors ModelBuilder instantiation to occur earlier in the local runner flow


sainivedh (approved):
lgtm

PR #728, opened 2025-08-01
pipeline_steps should be used in templates

zeiler:
We should be consistent with our API in using pipeline_steps not pipeline-steps in the templates. 

This make upload require some steps to upload for now so avoid confusion until lock is implemented. 

This updates the version IDs for existing template that have versions to make sure the config.yaml is always the final one uploaded. This caused a hard to understand bug about step name not being defined from the API when it had an old version ID in the config.yaml (which I guess was also in the uploaded API representation from pl upload).

Also improved logging by using proto to dict.

### DIFF
--- a/clarifai/cli/templates/pipeline_templates.py
+++ b/clarifai/cli/templates/pipeline_templates.py
@@ -25,12 +25,12 @@ def get_pipeline_config_template():
           steps:
           - - name: step-a
               templateRef:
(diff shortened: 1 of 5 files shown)


patricklundquist (approved):

zeiler:
@copilot fix the tests please

nitinbhojwani (approved):

PR #736, opened 2025-08-08
[PR-734] Use Method signature for local-runner (#718)

ackizilkale:
### Why

Temperature rather than 1.0 has been deprecated for gpt-5 models


luv-bansal (approved):
Seems to be wrong title name, but approving it

PR #739, opened 2025-08-12
[PR-768]: Fix Model Upload Deployment

srikanthbachala20:
This pull request makes a minor update to the deployment setup flow for models. The most notable change is updating the URL used to create new deployments, ensuring users are directed to the correct location in the UI.

Deployment flow update:

* Changed the deployment creation URL in `setup_deployment_for_model` to point to `/compute/deployments/create` instead of `/settings/compute/deployments/new`, ensuring users are sent to the correct page for creating deployments.
* Removed the log line that tells users they can return to the same deployment creation URL, as it is now redundant.


mogith-pn (approved):
LGTM

ackizilkale (approved):

PR #742, opened 2025-08-13
[Pr-766] add stream_options validation for internal streaming model upload 

christineyu123:
### PR Overview
Purpose: Validates that Clarifai internal models using OpenAI chat completions for streaming call (i.e. generate()) include required stream_options={'include_usage': True}

Key Logic:
    - Step 1Only applies validation to Clarifai internal users (via _is_clarifai_internal())
    - Step 2 Scans Python code for `chat.completions.create` and `generate( ` patterns
    - Step 3Ensures 'include_usage': True is present in stream_options

Notes to reviewers: Please verify if above key logic is good, the **challenge is in Step 2 where we need to find good conditions to know whether we need to trigger** the Step 3 validation. Please suggest if there are other ideas.

TODO: @christineyu123 before merging, perform some testing to make sure not hinder any current model upload behavior


clarifai/runners/models/model_builder.py
```diff
@@ -440,6 +441,89 @@ def _validate_config(self):
...
+        for root, _, files in os.walk(model_folder):
+            for file in files:
+                if file.endswith('.py'):
+                    file_path = os.path.join(root, file)
+                    try:
+                        with open(file_path, 'r', encoding='utf-8') as f:
+                            all_content += f.read() + "\n"
+                    except Exception:
+                        continue
+        return all_content
```
mogith-pn, line 515:
```suggestion
                           all_content.append(f.read())
                    except Exception:
                        continue
        return "\n".join(all_content)
```

christineyu123:
@mogith-pn 
1. Yes, I just updated the PR to use all_content as list. Thanks for the suggestion
2. For the verification implementation, I gave it more thoughts and discussed with Arman: we want to do minimum and super simple verification here to act as a sanity check without adding complexity to the code or slow down the upload process. 

Hence, the best way might be: 
1. I would keep the verification as simple as possible, as-is for now ( as long as there is include_usage and set_context in the string, we are good)
2. At same time, we reached out people to see if there will be warning at the pricing side; also, we can think of more general refactoring in future to force the implementation of token count via inheritance.

Does it make sense? Let me know your thought ;) thanks!

PR #765, opened 2025-08-28
Fix Local Runner CLI command

srikanthbachala20:
This pull request refactors the local model runner logic in `clarifai/cli/model.py` to improve clarity and modularity. The main change is replacing the direct use of the `serve` function with the `ModelServer` class, allowing for better encapsulation of server-related logic.

**Refactoring model serving logic:**

* Replaced the import of the `serve` function from `clarifai.runners.server` with an import of the `ModelServer` class, and updated the runner to instantiate a `ModelServer` object instead of calling the `serve` function directly.
* Updated the code to create a `ModelServer` instance and invoke its `serve` method, rather than calling the standalone `serve` function, for starting the local runner.

### DIFF
--- a/clarifai/cli/model.py
+++ b/clarifai/cli/model.py
@@ -569,9 +569,7 @@ def local_runner(ctx, model_path, pool_size, verbose):
     """
     from clarifai.client.user import User
     from clarifai.runners.models.model_builder import ModelBuilder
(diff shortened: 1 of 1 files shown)


phatvo9 (approved):
lgtm

PR #777, opened 2025-09-09
[EAGLE-6656]:  Use latest local-dev model version in local runner

patricklundquist:
The `clarifai model local-runner` command attempts to patch the latest existing version. If patching fails, or if no versions exist at all, a new version is created with the `local_dev` flag set to `True`.


luv-bansal (approved):
Looks good!

PR #822, opened 2025-10-22
[EAGLE-5341] Load model proto once to stop expecting it from predict request

wemoveon2:
### What

* Currently, the API sends the full model proto as part of each predict request

### Why

* This is unnecessary as we can load the model proto once and reuse it across all requests without depending on the API including it

### How

* Retrieve the model proto once via GetModel and set it in each predict request incase it's needed downstream

### Tests

* Existing tests pass

### Notes

*

### DIFF
--- a/tests/cli/test_pipeline.py
+++ b/tests/cli/test_pipeline.py
@@ -1445,7 +1445,7 @@ class TestPipelineListCommand:
         # Import here to avoid circular imports in testing
         from clarifai.cli.pipeline import list as list_command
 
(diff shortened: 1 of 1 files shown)


luv-bansal (approved):
Looks good

clarifai/runners/server.py
zeiler: isn't this path missing the model proto since it's only added to the ModelRunner. How come we wouldn't GetModel in the self._current_model so it's proto and model class are always together regardless of how they are served. @luv-bansal @wemoveon2

PR #824, opened 2025-10-24
Add 'visual-keypointer' to model types list

sanjaychelliah:
### Why

* Uploading visual-keypointer with concepts is running into error

### How

* Adding visual-keypointer in const.py


srikanthbachala20 (approved):
LGTM

PR #835, opened 2025-11-03
[PR-941]: Fixed the Error-Handling for Non-Existent Models

srikanthbachala20:
This pull request introduces a validation mechanism to the model loading process in `clarifai/client/model.py`, improving reliability during model initialization. The most important changes are:

**Model Initialization Enhancements:**

* The `__init__` method now calls `self.load_info(validate=True)` to ensure model information is validated when an instance is created.

**Model Info Loading Improvements:**

* The `load_info` method signature is updated to accept a `validate` boolean parameter, allowing for conditional validation during the model info loading process.
* The logic inside `load_info` is modified so that if `validate` is `False`, it processes and stores model information as before; otherwise, it skips this step, focusing on validation.

### DIFF
--- a/clarifai/client/model.py
+++ b/clarifai/client/model.py
@@ -1238,6 +1238,19 @@ class Model(Lister, BaseClient):
         return [concept_info['concept_id'] for concept_info in all_concepts_infos]
 
     def load_info(self, validate: bool = False) -> None:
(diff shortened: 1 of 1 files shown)


mogith-pn (approved):
Just made one comment.

PR #895, opened 2026-01-07
Backtrack & cleanup ambiguity removed

brajrajnagar:
### What changed
Backtrack & cleanup now runs only when deployment fails. It no longer runs after successful deployments.

### Why
Previously, Right after successful deployment user is asked "Do you want to backtrack and clean up?" Which is ambiguous and could lead to unintentionally deletion of deployments or model versions.

### New behavior
- Cleanup runs only on deployment failure
- No cleanup or prompt on successful deployment

### Testing
- Verified successful deployment does not trigger cleanup
- Verified failed deployment triggers backtrack & cleanup

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 clarifai-grpc>=12.0.2
-clarifai-protocol>=0.0.33
+clarifai-protocol==0.0.35
 numpy>=1.22.0
 tqdm>=4.65.0
 PyYAML>=6.0.1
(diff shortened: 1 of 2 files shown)


ackizilkale:
cause looking at the code I do not understand which change disables backtrack path for successful deployment

brajrajnagar:
@ackizilkale from line 2050 to 2057 I commented It gives users backtrack and cleanup option after deployment Which is confusing for user .
from line 2039 to 2048 I added backtrack_workflow to cleanup failed uploads and deployments.
It tested as expected for successful and failed deployment.

PR #916, opened 2026-02-03
[EAGLE-7160]: Update BaseClient to use init process

patricklundquist:
Adds init process used to reap zombies that might get created by terminated subprocesses. Uses spawn for CUDA compatibility. Enabled by default in Docker templates.


patricklundquist:
missed that BaseRunner doesn't call super().__init__

PR #994, opened 2026-03-18
perf: optimize model runner memory and latency

patricklundquist:
1. Fix memory leak / blocking bug in `runner_item_stream` by using `next()` and `itertools.chain()` instead of `list()`.
2. Eliminate list allocations for status checks.
3. Mutate status messages in place instead of allocating new ones.
4. Optimize _convert_input_protos_to_python in ModelClass


luv-bansal (approved):
Looks Good!

PR #1007, opened 2026-04-02
feat(runner): redirect local runner stdout/stderr to fluentd

patricklundquist:
When running a local runner inside a developer pod, standard output and standard error are redirected to `/proc/1/fd/1` using `os.dup2`. This allows fluentd to capture and route the local runner's logs as part of the main pod logs. This behavior is enabled when the `CLARIFAI_DEV_POD` environment variable is set to `"true"`.


patricklundquist:
closing, better to just use bash redirect and tee, e.g. `clarifai model local-runner ... 2>&1 | tee -a /var/log/dev_pod_logs.log`

PR #1012, opened 2026-04-08
`clarifai pipeline run --dev`

rizzip:
### Why

* Allow simpler testing for pipeline development, isolated from mainline pipeline objects but leveraging full orchestration rather than local execution

### How

* Added `--dev` flag to `clarifai pipeline run` to upload local code to an ephemeral `{pipeline_id}-dev` pipeline before running; only changed steps are re-uploaded using hash-based change detection
* Added optional positional `PATH` argument to `clarifai pipeline run` for specifying a pipeline directory or config file (prefers `config-lock.yaml`, falls back to `config.yaml`)
* Cleaned up logs for `clarifai pipeline run`

### Tests

*

### Notes

* https://clarifai.atlassian.net/browse/PR-1346

### DIFF
--- a/clarifai/cli/pipeline.py
+++ b/clarifai/cli/pipeline.py
@@ -297,17 +297,19 @@ def run(
                 argo_args_override=resources_pb2.ArgoArgsOverride(parameters=parameters)
             )
 
-    if monitor:
-        # Monitor existing pipeline run instead of starting new one
-        result = pipeline.monitor_only(timeout=timeout, monitor_interval=monitor_interval)
-    else:
-        # Start new pipeline run and monitor it
-        result = pipeline.run(
-            timeout=timeout,
-            monitor_interval=monitor_interval,
-            input_args_override=input_args_override,
-        )
-    click.echo(json.dumps(result, indent=2, default=str))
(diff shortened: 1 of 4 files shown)


clarifai/cli/pipeline.py
```diff
@@ -375,17 +413,19 @@ def run(
...
-        # Monitor existing pipeline run instead of starting new one
-        result = pipeline.monitor_only(timeout=timeout, monitor_interval=monitor_interval)
-    else:
-        # Start new pipeline run and monitor it
-        result = pipeline.run(
-            timeout=timeout,
-            monitor_interval=monitor_interval,
-            input_args_override=input_args_override,
-        )
-    click.echo(json.dumps(result, indent=2, default=str))
```
ackizilkale, line 388: did you remove this intentionally?

ackizilkale (approved):
lgtm. I was just curious about one log you removed. If not important all good.

clarifai/cli/pipeline.py
rizzip: Yes, the json was polluting the output imo.

PR #1013, opened 2026-04-09
Add clarifai pipeline local-run to run pipeline steps locally in Docker 

christineyu123:
Adds `clarifai pipeline local-run` command to run a pipeline step's pipeline_step.py locally in a Docker container. Minimum implementation that reuses existing `model serve --mode container` implementation pattern.
                                                                                                                                                                
  1. Only supports `--mode container` (default, no other modes). Usage: `clarifai pipeline local-run ./my-step` or `clarifai pipeline local-run ./my-step --mode container`
  2. Points at a specific pipeline step directory (containing config.yaml, requirements.txt, 1/pipeline_step.py), builds a Docker image, runs pipeline_step.py with all default values 
  3. Further pipeline local testing including end-to-end Argo workflow integration is not considered for this PR. We will consider in future if there is real need.

### DIFF
--- a/tests/runners/dummy_pipeline_step/1/pipeline_step.py
+++ b/tests/runners/dummy_pipeline_step/1/pipeline_step.py
@@ -1,0 +1,4 @@
+"""Dummy pipeline step for testing local-run."""
+
+print("Pipeline step started")
+print("Pipeline step completed successfully")
--- a/tests/runners/dummy_pipeline_step/1/pipeline_step.py
+++ b/tests/runners/dummy_pipeline_step/1/pipeline_step.py
@@ -1,17 +1,18 @@
-"""Dummy pipeline step matching the default clarifai pipelinestep init template."""
+"""Dummy pipeline step matching the default from clarifai pipeline init."""
 
 import argparse
 
+from clarifai.utils.logging import logger
+
 
 def main():
-    parser = argparse.ArgumentParser(description='Dummy pipeline step.')
-    parser.add_argument('--param_a', type=str, required=True, help='First parameter')
-    parser.add_argument('--param_b', type=str, default='default_b', help='Second parameter')
+    parser = argparse.ArgumentParser(description='dummy-pipeline-step processing step.')
+    parser.add_argument('--input_text', type=str, required=True, help='Text input for processing')
 
     args = parser.parse_args()
 
-    print(f"Pipeline step started: param_a={args.param_a}, param_b={args.param_b}")
-    print("Pipeline step completed successfully")
+    # TODO: Implement your pipeline step logic here
+    logger.info(f"dummy-pipeline-step processed: {args.input_text}")
 
 
 if __name__ == "__main__":
--- a/tests/runners/dummy_pipeline_step/1/pipeline_step.py
+++ b/tests/runners/dummy_pipeline_step/1/pipeline_step.py
@@ -1,4 +1,18 @@
-"""Dummy pipeline step for testing local-run."""
+"""Dummy pipeline step matching the default clarifai pipelinestep init template."""
 
-print("Pipeline step started")
-print("Pipeline step completed successfully")
+import argparse
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Dummy pipeline step.')
+    parser.add_argument('--param_a', type=str, required=True, help='First parameter')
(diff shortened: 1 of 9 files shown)


rizzip:
@christineyu123 thoughts re: `clarifai pipeline local-run` vs `clarifai pipeline run --local`? Additionally, as this is only supports running of a pipeline step, should this be a subcommand of `pipeline` or `pipelinestep`?

phatvo9:
@christineyu123 Just curious if we can specify GPU IDs in CLI? e.g., 0–1 are busy, I want to use 2–3 and I don't want to change python code.

christineyu123:
> @christineyu123 thoughts re: `clarifai pipeline local-run` vs `clarifai pipeline run --local`? Additionally, as this is only supports running of a pipeline step, should this be a subcommand of `pipeline` or `pipelinestep`?

@rizzip I was hesitating with local-run vs --local, but to be more consistent with updated model CLI, we use either `clarifai model local-runner` or `clarifai model serve` for running the model locally, I guess clarifai pipeline local-run is ok?

Also, I was thinking if worth to use switch pipelinestep, but since most of the CLI is starting with `clarifai pipeline xxx` I feel it is ok to keep it there for easier maintenance and retrieval. 

What do you think?

christineyu123:
> @christineyu123 Just curious if we can specify GPU IDs in CLI? e.g., 0–1 are busy, I want to use 2–3 and I don't want to change python code.

I did not add any ENV or extra flags in this PR: 1) I was trying to limit the scope without too many other flags 2) From discission with Nitin, if in future we want to have a more "integrated" local runner version including the full parameter passing and env setup, we should do it with argoworkflow simulation e.g. passing them e.g. CUDA devices as ENV as if we are launching real pipeline pods. Hence, I did not include this customization in this PR.

Is it Ok?

rizzip:
@christineyu123 

> @rizzip I was hesitating with local-run vs --local, but to be more consistent with updated model CLI, we use either clarifai model local-runner or clarifai model serve for running the model locally, I guess clarifai pipeline local-run is ok?

I'm sure it works, it's just semantically off since `clarifai pipeline ...` usually handles the pipelines objects, which can be multi-step, but this `local-run` command doesn't run the pipeline, but rather a single step. If the `clarifai pipeline ...` commands accept a pipeline directory / config, then `clarifai pipeline local-run` is inconsistent.

> Also, I was thinking if worth to use switch pipelinestep, but since most of the CLI is starting with clarifai pipeline xxx I feel it is ok to keep it there for easier maintenance and retrieval.

I think there was some thought to flattening the command structure, but I would think that a local run of a pipeline step should be attached to `clarifai pipelinestep` because that is the object under test.

christineyu123:
> @christineyu123
> 
> > @rizzip I was hesitating with local-run vs --local, but to be more consistent with updated model CLI, we use either clarifai model local-runner or clarifai model serve for running the model locally, I guess clarifai pipeline local-run is ok?
> 
> I'm sure it works, it's just semantically off since `clarifai pipeline ...` usually handles the pipelines objects, which can be multi-step, but this `local-run` command doesn't run the pipeline, but rather a single step. If the `clarifai pipeline ...` commands accept a pipeline directory / config, then `clarifai pipeline local-run` is inconsistent.
> 
> > Also, I was thinking if worth to use switch pipelinestep, but since most of the CLI is starting with clarifai pipeline xxx I feel it is ok to keep it there for easier maintenance and retrieval.
> 
> I think there was some thought to flattening the command structure, but I would think that a local run of a pipeline step should be attached to `clarifai pipelinestep` because that is the object under test.

Yes, after some more thoughts and the pipeline's objects issues are definitely valid. I will make refactoring and let you know once fixed with  `clarifai pipelinestep`! Thanks

tests/runners/dummy_pipeline_step/1/pipeline_step.py
rizzip, line 1: This file does not resemble the default from `clarifai pipeline init` (no arguments at all). The default template has 1 required arg, which fails to run locally, and any required arguments will cause a failure.

rizzip (changes requested):
@christineyu123 local run does not work with the default from `clarifai pipeline init` due to a required argument. Recommend updating the test cases and allowing argument passthrough (e.g., `clarifai pipelinestep local-run PATH -- STEP_ARGS`)

rizzip (approved):

PR #960, opened 2026-02-21
CLI improvement

luv-bansal:
## Pull request overview

This PR modernizes and expands the Clarifai CLI experience around auth/config management and model workflows (init/serve/upload/deploy), while adding offline-friendly embedded toolkit templates and substantially expanding test coverage for the updated CLI flows.

PR #1018, opened 2026-04-20
validate private HuggingFace repos with token   

christineyu123:
The two-step HF validation added in #960 (2026-03-04) does an anonymous probe first, then retries with the  
  user's token only when the reason is "gated_no_token". But HuggingFace returns RepositoryNotFoundError →
  "not_found" for private repos to anonymous callers (to avoid leaking their existence), so private repos     
  never got the retry and failed with a misleading "not found" error — even when the user had a valid token
  with access.                                                                                                
                                                                                                              
  This PR expands the retry branch to also handle "not_found", so private HF repos resolve correctly via      
  HF_TOKEN or checkpoints.hf_token. Truly missing repos still raise the "not found" error once the
  token-authenticated probe also fails. Updated one existing test and added a new                             
  test_validate_config_private_repo_retries_with_token to cover the private-repo path.

### DIFF
--- a/clarifai/runners/models/model_builder.py
+++ b/clarifai/runners/models/model_builder.py
@@ -692,8 +692,9 @@ class ModelBuilder:
                 if has_access:
                     # Public repo — no token needed anywhere.
                     pass
-                elif reason == "gated_no_token":
-                    # Repo requires auth. Validate with the available token.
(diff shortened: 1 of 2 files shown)


luv-bansal (approved):
Looks good

PR #1019, opened 2026-04-22
[PR-1309] add sglang cancellation

phatvo9:
### What

Add sglang openai class + cancellation

### Tests
* Manual tests via local runner
* Mock tests


christineyu123:
Looks good! What I would suggest is if you can find a model that you are currently hardcoding the cancellation, and try replace model.py using this PR's model class and see if def generate()/openai_stream_support(), etc. are still working. Especially the token count and pricing are still good. To ensure nothing wrong if future models start to inherit that .

christineyu123 (approved):

PR #1024, opened 2026-04-27
Re-pin deployment desired_worker to current model_version on serve

ackizilkale:
When `clarifai model serve` reuses an existing deployment, the patch silently failed for two reasons:

  1. PatchDeployments only accepts action='overwrite' — 'merge' is rejected by the backend ("Invalid action: Unrecognized action 'merge'. Supported actions: overwrite"). The CLI used 'merge', and the failure was swallowed by `except Exception: pass`, so even the visibility patch this code claimed to do was never landing.

  2. `Deployment.worker` is the *observed* state (read-only on input); writes to it are silently ignored by the backend. The writable input is `Deployment.desired_worker`.

With deploy_latest_version=False (set in #1022 for serve deployments), the deployment is pinned to whatever desired_worker.model.model_version.id it has. When the method-signatures hash changes a new model_version and runner are created, but the deployment's desired_worker stayed on the old version_id — leaving API calls routed to a version with no live runner ("Model is still deploying...").

Switch action to 'overwrite', patch desired_worker (not worker), and surface non-success responses + exceptions as warnings instead of swallowing them. Idempotent when the version is unchanged; heals the binding when it changed.

### DIFF
--- a/clarifai/cli/model.py
+++ b/clarifai/cli/model.py
@@ -2265,9 +2265,7 @@ def serve_cmd(
                     ],
                     action='overwrite',
                 )
-                resp = nodepool._grpc_request(
-                    nodepool.STUB.PatchDeployments, patch_req
-                )
+                resp = nodepool._grpc_request(nodepool.STUB.PatchDeployments, patch_req)
                 if resp.status.code != status_code_pb2.SUCCESS:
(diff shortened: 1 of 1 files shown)


luv-bansal (approved):
LGTM

PR #1044, opened 2026-05-20
Enable base image and platform specification at the DSL level

rizzip:
### Why

* Pipeline DSL builder uses default docker image with only python installed, making use fo CUDA optimized images difficult
* Build process defaults to building for both amd64 and arm64, which is not always preferred

### How

* Adds `base_image` and `platform` kwargs to `@step`
* Installs and uses `uv` for requirements resolution when using the built in Dockerfile

### Tests

* Relevent unit tests added
* [Sister PR in pipeline-examples](https://github.com/Clarifai/pipeline-examples/pull/33) shows example usage

### Notes

*


christineyu123 (approved):

clarifai/cli/pipeline.py
```diff
@@ -128,16 +142,43 @@ def upload(path, no_lockfile):
...
     pipeline_obj = load_pipeline_from_file(path)
+    if user_id:
+        pipeline_obj.user_id = user_id
+    if app_id:
+        pipeline_obj.app_id = app_id
     config_path = pipeline_obj.generate(output_dir)
-    logger.info(f"Generated pipeline assets at {config_path}")
+
+    # Generate Dockerfiles for all locally managed step directories.
+    # Note: avoid calling list() here — `list` is shadowed by a Click command in this module.
```
nitinbhojwani: Good catch. To remove this comment, how about fixing the existing method name from list to `list_pipelines` (since it's not a good practice to shadow a built-in type) and explicitly name that via decorator e.g. `@pipeline.command(name='list', aliases=['ls'])` ?

rizzip: @copilot implement this fix to avoid shadowing the builtin

PR #1049, opened 2026-05-25
Add admission control to VLLMOpenAIModelClass

christineyu123:
Move the kv_cache-based admission control (`VLLMMetricsPoller` background poller + three-state deadband
  `check_admission`) from per-model implementations into the base `VLLMOpenAIModelClass`. Subclasses now only need to   
  instantiate `self._metrics_poller = VLLMMetricsPoller(base_url, poll_interval=5.0)` inside `load_model()` to opt in — 
  the parent provides `check_admission` (returns `False` above `KV_CACHE_HIGH=0.8`, `True` below `KV_CACHE_LOW=0.5`,   
  `None` in between) plus `admission_increase_delay`/`admission_decrease_delay` properties. Fails open when the poller  
  is unset or stale, so subclasses that don't initialize it run unchanged.


clarifai/runners/models/vllm_openai_class.py
```diff
@@ -1,10 +1,65 @@
...
+
+class VLLMMetricsPoller:
+    """Polls vLLM /metrics in background; caches kv_cache usage.
+
+    Fail-open: if the poller has never succeeded or is stale, admission is allowed.
+    """
+
+    KV_CACHE_HIGH = 0.8
+    KV_CACHE_LOW = 0.5
+    STALE_AFTER_SECONDS = 5.0
```
ackizilkale, line 21: Because these are class constants we can't modify them at instance level. Do you think we should put these under __init__ so can be modified per instance?

ackizilkale (approved):
minor comment.

